PR #1, opened 2023-05-31
Add support for the L56 discovery board

loqoman:
Adding support for the STM32L562 discovery board, as an (expensive) alternative to the nucleo for remote work. 

**Changelog**
- Add `ZeroPilot-SW-3.5/boardfiles/discoveryl562qe`
    - All peripherals in `discoveryl562qe.ioc` are defined, as there are almost no user GPIO pins exposed on the board
    - Basic HAL methods have been confirmed to work (`HAL_GPIO_TogglePin()` and `HAL_Delay()`)
    - All peripheral init methods have been commented out in `main.c`, as `MX_SDMMC1_SD_Init()` was faulting. A later infra task I will take on is to further chase out what does and doesn't work, however it's out to scope for this PR.


Chrisc110:
Seeing as though you are just adding the boardfiles and not modifying the build script looks good to me… since the only reason you are doing this is bec the discovery board is the only board you have, we may end up removing this later down the line just bec it isn’t a board that we would ever use for flight or testing. Other than that, looks good to me.

Chrisc110 (approved):

loqoman:
Totally agree @Chrisc110 . I saw the build script has support for a parameter for the compile target, and I'll be using that. So long as I'll develop against the discovery board I have no problem fronting the work to maintain it. shouldn't impede on any other dev work

SnackkOverflowError (approved):
lgtm

loqoman:
Merging.

PR #5, opened 2023-06-13
toggle on the execute bit for test/build scripts

Aidan-B:
## Description

Changed the mode, adding the `+x` (execute) bit for use of the scripts on linux systems.

## Testing

Both scripts were run on my laptop, but should be tested on windows before merging.


Aidan-B:
Will be addressed in #9

PR #10, opened 2023-06-16
sensor fusion added

HardyYu:
### About the changes I made
1. added the sensor fusion code and all the associated files such as some common defines and math files from old ZP to ZP3.5
2. implement a **SF_Config.h** file which has the defines that coordinate with milestone3 and milestone4. When milestone 3 is completed, the **SF_milestone3** can be defined, and the imu part of the sensor fusion will be activated. When milestone 4 is done, all of the sensor fusions are activated. This design hopefully would help ppl on using SF and testing with SF
3. Added an additional function **SFError_t SensorFusionInterfaceOutput(SFOutput_t (asterisk)SFoutput)** that is not on the old SF, which serves for fetching the output of sensor fusion purpose
4.  changed CMake such that the program could build

### Why this is threadsafe
1. Basically, the sensor fusion is the **SF_Output** producer, and AM/PM are the **SF_Output** consumer
2. There is only a single thread that would change **SF_Output** which is the sensor fusion thread
3. Despite that **SF_Output** has multiple consumers, mutex or semaphores are not needed because only one thread can change the value
4. As long as the ppl who use the code keep good hygiene on how this thread runs, there shouldn't be race condition happening

### Things Tested
1. Running in multi-thread along with some fake threads by FreeRTOS in a testing project
2. After porting the code to ZP3.5 I build the code with bash
3. Run some functions in the main while loop

## Documentation

Please let me know if any things should be changed, such as the coding format, file structure, or potential coding issues.


Drivers/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -0,0 +1,604 @@
...
+SFError_t SF_GetResult(SFOutput_t *output)
+{
+    SFError_t SFError;
+
+    //TODO: This function always return success as return, fix it?
+    SFError.errorCode = 0;
+
+    *output = SFOutput;
+
+    return SFError;
```
Aidan-B, line 551: This does not seem thread safe. It simply returns a static pointer to the data that we put the updated data in.

Aidan-B (changes requested):
I think the consumer-producer model for thread safety only works if there is a queue between the consumer and producer, which is not the scenario we have here. I think the interfacing mechanism needs to be thought through a little more.

Drivers/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+
+SFError_t SF_GetResult(SFOutput_t *output)
+{
+    SFError_t SFError;
+
+    //TODO: This function always return success as return, fix it?
+    SFError.errorCode = 0;
+
+    *output = SFOutput;
```
Aidan-B, line 549: Based on the discussion we had today, this should now be returned (by copy) as a static method with static member variables. It should be protected by a mutex provided by freertos for both reads and writes to the variable.

Drivers/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -534,19 +537,22 @@ SFError_t SF_GenerateNewResult()
...
 {
     SFError_t SFError;
 
     //TODO: This function always return success as return, fix it?
     SFError.errorCode = 0;
 
-    *output = SFOutput;
+    xSemaphoreTake(SF_mutex, portMAX_DELAY);
+
+    output = &SFOutput;
```
Aidan-B, line 553: A mutex wouldn't be required for this operation, since you're providing a reference to the output. We really should be doing a copy here instead so that the caller will get it's own copy of the data that it can safely interact with.

Drivers/SensorFusion/Src/sensorFusionInterface.cpp
```diff
@@ -1,7 +1,7 @@
 #include "sensorFusionInterface.hpp"
 #include "SensorFusion.hpp"
 
-void SensorFusionInterfaceInit(void) {
+bool SensorFusionInterfaceInit(void) {
     SF_Init();
 }
```
Aidan-B: This doesn't have a return value

build/CMakeCache.txt
Aidan-B, line 1: You appear to have committed your build folder & a bunch of IDE settings. These should probably not be committed.

Aidan-B (changes requested):

Drivers/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -534,19 +537,22 @@ SFError_t SF_GenerateNewResult()
     #endif
 
 #endif
+    xSemaphoreGive(SF_mutex);
```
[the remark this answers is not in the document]
Aidan-B: i.e. copy the data you calculate into the SFOutput only when you're done all calculations, and only grab the mutex when you do that copy

Boardfiles/nucleol552zeq/Core/Src/main.c
Aidan-B, line 1: Any reason this was converted to `.c`?

Drivers/SensorFusion/Src/MahonyAHRS.cpp
```diff
@@ -63,15 +63,15 @@ void MahonyAHRSupdate(float gx, float gy, float gz, float ax, float ay, float az
 
 		// Normalise accelerometer measurement
 		recipNorm = invSqrt(ax * ax + ay * ay + az * az);
-		ax *= recipNorm;
-		ay *= recipNorm;
-		az *= recipNorm;     
+		ax = ax * recipNorm;
+		ay = ay * recipNorm;
+		az = az * recipNorm;
```
Aidan-B: I'm curious if you know what exactly the impact is that this has on the system and why the previous operation was giving a build warning? I didn't look to far into it.

Drivers/SensorFusion/Src/sensor_fusion.cpp
```diff
@@ -488,23 +495,22 @@ SFError_t SF_GenerateNewResult()
 {
     SFError_t SFError;
     SFError.errorCode = 0;
-    xSemaphoreTake(SF_mutex, portMAX_DELAY);
+    SFAttitudeOutput_t attitudeOutput;
+    memset(&attitudeOutput, 0, sizeof(attitudeOutput));
+    SFPathOutput_t pathOutput;
+    memset(&pathOutput, 0, sizeof(pathOutput));
```
Aidan-B, line 501: Why is this `memset` needed?

Boardfiles/nucleol552zeq/.settings/language.settings.xml
Aidan-B, line 1: I think you missed removing this one

Drivers/SensorFusion/Inc/sensor_fusion.hpp
```diff
@@ -14,8 +14,10 @@ extern "C"
 #ifndef SENSORFUSION_HPP
 #define SENSORFUSION_HPP
 
-#include "CommonDataTypes.hpp"
-#include "SF_Config.h"
+#include "common_data_types.hpp"
+#include "SF_config.h"
+#include "FreeRTOS.h"
+#include "semphr.h"
```
Aidan-B, line 20: Are these used in this file? We should only include the files we use in the file they are used in.

Drivers/SensorFusion/Src/sensor_fusion.cpp
```diff
@@ -51,6 +51,9 @@ struct SFIterationData_t{
     float prevP[NUM_KALMAN_VALUES*NUM_KALMAN_VALUES];
 };
 
+// Globals
+static SemaphoreHandle_t SF_mutex;
```
Aidan-B, line 55: Any particular reason this needs to be global?

Aidan-B (changes requested):

Boardfiles/nucleol552zeq/Core/Src/main.c
HardyYu: I thought the original main is .c, so I was trying to keep the consistency

Drivers/SensorFusion/Src/MahonyAHRS.cpp
HardyYu: seems like += works differently from = then +. it is unsafe to actually use += for volatile values for some reason

Drivers/SensorFusion/Src/sensor_fusion.cpp
HardyYu: to not assign uninitiated value into other variables. to eliminate the warning that prompt from compile

Boardfiles/nucleol552zeq/.settings/language.settings.xml
HardyYu: will remove it

Drivers/SensorFusion/Src/sensor_fusion.cpp
HardyYu: global mutex implies that the same mutex or the lock can exist and work for different functions(read & write). that is what i thought

Aidan-B: My bad. I thought this was a C++ class

Drivers/SensorFusion/Inc/sensor_fusion.hpp
HardyYu: Good point. I tested today. Both libraries have to be there because SemaphoreHandle_t is used in sensor_fusion.cpp.

Drivers/SensorFusion/Inc/sensor_fusion.hpp
HardyYu, line 1: moved the include to the proper place

Aidan-B (approved):

Drivers/SensorFusion/Inc/common_data_types.hpp
```diff
@@ -0,0 +1,9 @@
+#ifndef COMMON_DATA_TYPES_HPP
+#define COMMON_DATA_TYPES_HPP
+
+enum SensorErrorCodes {SENSOR_SUCCESS = 0, SENSOR_FAIL = 1};
```
SnackkOverflowError, line 4: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards style

Drivers/SensorFusion/Inc/sensor_fusion.hpp
```diff
@@ -0,0 +1,137 @@
...
+    bool isDataNew; 
+    enum SensorErrorCodes status; //TBD but probably 0 = SUCCESS, -1 = FAIL, 1 = BUSY 
+    int utcTime; //Last time GetResult was called
+} Altimeter_Data_t;
+
+
+/**
+ * Initialize sensor fusion.
+ */ 
+SFError_t SF_Init(void);
```
SnackkOverflowError, line 99: camelcase https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards

Drivers/SensorFusion/Src/MahonyAHRS.cpp
```diff
@@ -0,0 +1,233 @@
...
+	q0 = q0 * recipNorm;
+	q1 = q1 * recipNorm;
+	q2 = q2 * recipNorm;
+	q3 = q3 * recipNorm;
+}
+
+//---------------------------------------------------------------------------------------------------
+// IMU algorithm update
+
+void MahonyAHRSupdateIMU(float gx, float gy, float gz, float ax, float ay, float az) {
```
SnackkOverflowError, line 151: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards method names

SnackkOverflowError (commented):
code looks good, file names are going to have to be renamed, function names too.
https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards

SnackkOverflowError (approved):
Please delete temp.txt before merging

HardyYu:
sensor fusion added

PR #14, opened 2023-06-30
PWM out driver

cehune:
## Description

PWM Output Driver for ZP 3.5. 
Tested LOS PWM out driver. Added functionality to read system clock, and auto set prescaler based on desired frequency. 
By default the desired frequency is 50hz.

Required Driver

<br/>

## Testing

Tested manually on Seleae Logic Analyzer, with an input clock frequency of 48 MHz. 

Comfirmed a constant~50hz frequency, with between ~5-10.25% duty cycle. (There is a small margin of error above and below 5 and 10%). 

Repeatedly input percent between 5 and 10%, confirming that it will reliably output the correct duty cycle.

<br/>

None so far, in progress.

I aim to test this with varying system clocks.

<br/>

## Documentation

*Milestone number and name:*

[*Link to Asana task:*](https://app.asana.com/0/home/1203900823953668/1204698388503807)

[*Link to Confluence documentation:*](https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2251817033/PWM+Out)

<br/>

## Reminders

- [x] Add reviewers to the PR

### DIFF
--- a/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
@@ -5,7 +5,6 @@
 
 class MotorChannel {
     public:
-        virtual void setup(void) = 0;
         virtual void set(uint8_t percent) = 0;
     protected:
         bool isSetup = false;
--- a/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
@@ -1,0 +1,14 @@
+#ifndef ZP_D_MOTOR_CHANNEL_HPP_
+#define ZP_D_MOTOR_CHANNEL_HPP_
+
+#include <cstdint>
+
+class MotorChannel {
+    public:
+        virtual void setup(void) = 0;
+        virtual void set(uint8_t percent) = 0;
+    protected:
+        bool isSetup = false;
+};
+
+#endif // ZP_D_MOTOR_CHANNEL_HPP_
--- a/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
@@ -6,8 +6,6 @@
 class MotorChannel {
     public:
         virtual void set(uint8_t percent) = 0;
-    protected:
-        bool isSetup = false;
 };
 
 #endif // ZP_D_MOTOR_CHANNEL_HPP_
--- a/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
@@ -15,12 +15,10 @@ class PWMChannel : public MotorChannel {
         TIM_HandleTypeDef *timer;
         const uint16_t TIMER_CHANNEL;
 
-    	//set to minimum and max amount of counts in a duty cycle
+    	// set to minimum and max amount of counts in a duty cycle
         const uint32_t MIN_SIGNAL;
         const uint32_t MAX_SIGNAL;
-
-        const uint32_t CLOCK_FREQUENCY = SystemCoreClock;    //system clock frequency
-        const uint16_t DESIRED_FREQUENCY = 50; //PWM frequency in hz
+        const uint16_t DESIRED_FREQUENCY = 50; // PWM frequency in hz
 };
 
 #endif // ZP_D_PWM_CHANNEL_HPP_
--- a/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
@@ -7,27 +7,20 @@
 
 class PWMChannel : public MotorChannel {
     public:
-        PWMChannel(uint16_t pin_num, GPIO_TypeDef* port, TIM_HandleTypeDef* timer,
-					uint16_t timer_channel,uint32_t clock_frequency);
+        PWMChannel(TIM_HandleTypeDef* timer,
+					uint16_t timer_channel);
 
-        void setup(void);
         void set(uint8_t percent);
     private:
-        uint16_t pin_num_;
-        GPIO_TypeDef* port_;
-        TIM_HandleTypeDef *timer_;     //handle to the timer
-        uint16_t timer_channel_;       //channel of the timer
-        uint32_t clock_frequency_;     //system clock frequency
-        uint32_t period_ticks_;
+    	//set to minimum and max amount of counts in a duty cycle
+        uint32_t MIN_SIGNAL;
+        uint32_t MAX_SIGNAL;
+        uint32_t PERIOD_TICKS_;  
+        TIM_HandleTypeDef *TIMER_;     //handle to the timer
+        const uint16_t TIMER_CHANNEL_;   //channel of the timer
 
-        //set to min and max amount of counts in a duty cycle
-        uint32_t min_signal;
-        uint32_t max_signal;
-
-        //values in microseconds(us)
-        const uint16_t DESIRED_FREQUENCY = 50;
-
-
+        const uint32_t CLOCK_FREQUENCY_ = SystemCoreClock;    //system clock frequency
+        const uint16_t DESIRED_FREQUENCY = 50; //PWM frequency in hz
 };
 
 #endif // ZP_D_PWM_CHANNEL_HPP_
--- a/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
+++ b/Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
@@ -1,0 +1,32 @@
+#ifndef ZP_D_PWM_CHANNEL_HPP_
+#define ZP_D_PWM_CHANNEL_HPP_
+
+#include "main.h"
+
+#include "ZP_D_MotorChannel.hpp"
+
+class PWMChannel : public MotorChannel {
+    public:
+        PWMChannel(uint16_t pin_num, GPIO_TypeDef* port, TIM_HandleTypeDef* timer,
+					uint16_t timer_channel,uint32_t clock_frequency);
+
+        void setup(void);
+        void set(uint8_t percent);
+    private:
+        uint16_t pin_num_;
+        GPIO_TypeDef* port_;
+        TIM_HandleTypeDef *timer_;     //handle to the timer
+        uint16_t timer_channel_;       //channel of the timer
+        uint32_t clock_frequency_;     //system clock frequency
+        uint32_t period_ticks_;
+
+        //values in microseconds(us)
+        static constexpr uint32_t PWM_PERIOD = 20000;
+        static constexpr uint32_t MIN_SIGNAL = 950; //standard for 50hz pwm signal, with a 100 Hz margin added to ensure we can reach the extreme values.
+        static constexpr uint32_t MAX_SIGNAL = 2050;
+        const uint16_t DESIRED_FREQUENCY = 50;
+
+
+};
+
+#endif // ZP_D_PWM_CHANNEL_HPP_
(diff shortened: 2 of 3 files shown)


Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,32 @@
...
+    private:
+        uint16_t pin_num_;
+        GPIO_TypeDef* port_;
+        TIM_HandleTypeDef *timer_;     //handle to the timer
+        uint16_t timer_channel_;       //channel of the timer
+        uint32_t clock_frequency_;     //system clock frequency
+        uint32_t period_ticks_;
+
+        //values in microseconds(us)
+        static constexpr uint32_t PWM_PERIOD = 20000;
```
Chrisc110, line 24: what is the difference between period_ticks_ and PWM_PERIOD? And why is PWM_PERIOD set to 20000?

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+
+void PWMChannel::set(uint8_t percent) {
+    if (percent > 100 || !isSetup) {
+        return;
+    }
+
+    uint32_t us = (percent * (MAX_SIGNAL - MIN_SIGNAL)) / 100 + MIN_SIGNAL;
+    uint32_t ticks = static_cast<uint32_t>((static_cast<float>(us) / static_cast<float>(PWM_PERIOD))
+                        * static_cast<float>(period_ticks_));
```
Chrisc110, line 31: could you explain what is going on in this calculation?

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
cehune: Sorry, I realize I should've looked closer at these. Since they were in the original LOS driver, I left them in.

I'm not sure how we should set the prescaler and period, as that would be two unknowns when all we have is the desired frequency and the system clock frequency. I assumed that PWM_PERIOD would just be our general counts amount. 

I will remove the redundancy, retest, and repush

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
cehune: Sorry again, this was another piece I didn't touch as it was in the LOS driver. Seems to just divide by PWM_PERIOD, then multiply by that same value. Will remove.

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
cehune: Or should it just take the counts period from whatever is already set on that timer?

cehune:
@Chrisc110 Could you recheck what I have?

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
Chrisc110: gotcha ok

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
Chrisc110: So your correct, if we want the feature where at startup, no matter the input frequency we can calculate the prescaler/period, that would be two unknowns. So that means we need to have one hardcoded, prop the period?? And at that point we would just solve for the prescaler and set both.

Chrisc110 (approved):

TheSpaceDragon (approved):

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
```diff
@@ -0,0 +1,42 @@
...
+	//Calculate new prescaler
+	uint16_t prescaler_ = (clock_frequency_ / DESIRED_FREQUENCY
+			/ period_ticks_) - 1;
+
+	__HAL_TIM_SET_PRESCALER(timer_, prescaler_);
+	HAL_TIM_PWM_Start(timer_, timer_channel_);
+
+
+	isSetup = true;
+}
```
Aidan-B, line 25: Any reason to have this "setup" function instead of just doing this work in the constructor?

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,33 @@
...
+class PWMChannel : public MotorChannel {
+    public:
+        PWMChannel(uint16_t pin_num, GPIO_TypeDef* port, TIM_HandleTypeDef* timer,
+					uint16_t timer_channel,uint32_t clock_frequency);
+
+        void setup(void);
+        void set(uint8_t percent);
+    private:
+        uint16_t pin_num_;
+        GPIO_TypeDef* port_;
```
Aidan-B, line 17: I'm pretty sure `pin_num_` and `port_` aren't actually used?

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef ZP_D_PWM_CHANNEL_HPP_
+#define ZP_D_PWM_CHANNEL_HPP_
+
+#include "main.h"
```
Aidan-B, line 4: Why is main included?

Aidan-B (changes requested):
Just a couple small things to address before I would approve

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,33 @@
...
+					uint16_t timer_channel,uint32_t clock_frequency);
+
+        void setup(void);
+        void set(uint8_t percent);
+    private:
+        uint16_t pin_num_;
+        GPIO_TypeDef* port_;
+        TIM_HandleTypeDef *timer_;     //handle to the timer
+        uint16_t timer_channel_;       //channel of the timer
+        uint32_t clock_frequency_;     //system clock frequency
```
Aidan-B, line 20: This is likely something we can determine at compile time and don't really need to initialize in the constructor is it?
A bunch of these can also be `const`

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
cehune: No, I can add it in there.

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
cehune: Yes, I'll remove these. I left them in because they were there from the LOS driver, and I didn't want to change the syntax

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,26 @@
...
+        PWMChannel(TIM_HandleTypeDef* timer,
+					uint16_t timer_channel);
+
+        void set(uint8_t percent);
+    private:
+    	//set to minimum and max amount of counts in a duty cycle
+        uint32_t MIN_SIGNAL;
+        uint32_t MAX_SIGNAL;
+        uint32_t PERIOD_TICKS_;  
+        TIM_HandleTypeDef *TIMER_;     //handle to the timer
```
Aidan-B, line 19: These are non `const`, so I'm pretty sure they shouldn't be all caps?
https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards#Member-Variables

Aidan-B (changes requested):
Last change. After this, I think it looks great! Great work! 🥳 

I think we also need to add unit tests, but I don't want to block this going in any further, so I would suggest adding unit tests in a new PR.

SnackkOverflowError (approved):
Please fix the style issues before merging, but code logic looks good

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,26 @@
...
+    public:
+        PWMChannel(TIM_HandleTypeDef* timer,
+					uint16_t timerChannel);
+
+        void set(uint8_t percent);
+    private:
+    	// set to minimum and max amount of counts in a duty cycle
+        uint32_t minSignal;
+        uint32_t maxSignal;
+        uint32_t PERIOD_TICKS;
```
Aidan-B, line 18: These variables aren't mutated by the class so we can mark them as `const` and bring back the all caps naming.
Additionally, `PERIOD_TICKS` doesn't need to be saved by the class because it isn't used, except in the constructor.
```suggestion
        const uint32_t MIN_SIGNAL;
        const uint32_t MAX_SIGNAL;
```

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
```diff
@@ -0,0 +1,39 @@
...
+	// Calculate new pre-scaler
+	uint16_t prescaler = (CLOCK_FREQUENCY / DESIRED_FREQUENCY
+			/ PERIOD_TICKS) - 1;
+
+	__HAL_TIM_SET_PRESCALER(TIMER, prescaler);
+	HAL_TIM_PWM_Start(TIMER, TIMER_CHANNEL);
+
+
+	isSetup = true;
+}
```
Aidan-B, line 23: We can use the initialization list for a whole bunch of these.
```suggestion
PWMChannel::PWMChannel(TIM_HandleTypeDef* timer,
                        uint16_t timerChannel) :
                        MIN_SIGNAL(TIMER->Init.Period * 0.05),  // sets counts for 5% duty cycle
                        MAX_SIGNAL(TIMER->Init.Period * 0.10),  // sets counts for 10% duty cycle
                        TIMER(timer),
                        TIMER_CHANNEL(timerChannel)
{
	// Calculate new pre-scaler
	uint16_t prescaler = (CLOCK_FREQUENCY / DESIRED_FREQUENCY
			/ TIMER->Init.Period) - 1;

	__HAL_TIM_SET_PRESCALER(TIMER, prescaler);
	HAL_TIM_PWM_Start(TIMER, TIMER_CHANNEL);
}
```

Drivers/MotorChannel/Src/ZP_D_PWMChannel.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+void PWMChannel::set(uint8_t percent) {
+	/* Sets the duty cycle as a percent between 5 and 10%.
+	 *
+	 * Usage:
+	 * 0% corresponds to a duty cycle of 5%
+	 * 100% corresponds to a duty cycle of 10%
+	 * 50% corresponds to a duty cycle of 7.5%*/
+    if (percent > 100 || !isSetup) {
```
Aidan-B, line 32: `isSetup` will always be true here, since we set it to be true in the constructor, so this check can be removed, and we can remove the `isSetup` variable all together.
```suggestion
    if (percent > 100) {
```

Drivers/MotorChannel/Inc/ZP_D_MotorChannel.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef ZP_D_MOTOR_CHANNEL_HPP_
+#define ZP_D_MOTOR_CHANNEL_HPP_
+
+#include <cstdint>
+
+class MotorChannel {
+    public:
+        virtual void set(uint8_t percent) = 0;
+    protected:
+        bool isSetup = false;
```
Aidan-B, line 10: This `isSetup` variable doesn't seem to have any utility the way the driver is currently structured, and should probably just be removed.

Drivers/MotorChannel/Inc/ZP_D_PWMChannel.hpp
```diff
@@ -0,0 +1,26 @@
...
+        TIM_HandleTypeDef *TIMER;     // handle to the timer
```
Aidan-B, line 19:
```suggestion
        TIM_HandleTypeDef *timer_;
```

Aidan-B (changes requested):

Aidan-B (approved):

PR #16, opened 2023-07-02
AM input interface

lekangwang:
## Description

- Added static getter, setter, and variable to AM class
- Defined controller inputs struct with 4 float fields
- Wrote unit tests for getter/setter
- Did some cmakelist.txt cleanup

<br/>

- To allow AM to receive controller inputs (roll, pitch, yaw, throttle) in a thread-safe way

<br/>

## Testing

- Firmware compiles successfully

<br/>

- Test using getter/setter to get/set the static variable belonging to the AM class

<br/>

## Documentation

*Milestone number and name:*
1. Manual fixed wing

*Link to Asana task:*
[https://app.asana.com/0/1203458353737758/1204799493473409/f](https://app.asana.com/0/1203458353737758/1204799493473409/f)

[https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2250768443/Attitude+Management#AM-Input](https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2250768443/Attitude+Management#AM-Input)

<br/>

## Reminders

- [x] Add reviewers to the PR


AttitudeManager/Tests/ExampleTest.cpp
Aidan-B, line 1: I think this generally looks good, although I have two open questions:
1. Should we make use of "gtest/gtest_prod.h", so that the private variables can be accessed so that only the function we want to test is in fact being tested? I know the general advice is to only test the interfaces of a class so I'm not sure.
2. Should we be testing that the synchronization of the mutexes works as expected? Can this be done? I haven't done much research on this before so I have no idea if it is possible, or best practive

AttitudeManager/Tests/ExampleTest.cpp
```diff
@@ -7,3 +8,33 @@ TEST(AttitudeManager, BasicAssertions) {
     // Expect equality.
     EXPECT_EQ(7 * 6, 42);
 }
```
Aidan-B, line 10: You can remove this example test

AttitudeManager/Src/AM.cpp
```diff
@@ -2,13 +2,36 @@
 // Created by Anthony Luo on 2022-10-12.
 //
 #include "AM.hpp"
+#include "semphr.h"
 
 #include <array>
 #include <cstdlib>
 
 namespace AM {
+AttitudeManager::control_inputs_mutex = xSemaphoreCreateMutex();
```
Aidan-B, line 11: Should this be done in the AttitudeManager constructor? This seems like an odd place to define the mutex.

Aidan-B (commented):
I think this looks great so far. I assume your intention is that whenever SM gets controller input it will pass it to the `setControlInputs()` method, and whenever AM executes, it will get instructions with the `getControlInputs()` method?

Also going to add @SnackkOverflowError to take a peek. He might have some insight around unit testing best practices here.

AttitudeManager/Tests/ExampleTest.cpp
lekangwang: Done

AttitudeManager/Tests/ExampleTest.cpp
lekangwang: Yes my intent is to use the getter/setter to interface with the static variables.

For your comment 2. I think testing the synchronization of the mutex would be a better fit with integration tests involving SM/AM rather than unit tests. 

This code is, in my opinion, simple enough to verify by eye that the mutex is protecting crit. sections properly.

AttitudeManager/Src/AM.cpp
lekangwang: I thought static variables aren't associated with a class instance? I don't think it would work to put these in the AM class constructor.

AttitudeManager/Inc/AM.hpp
```diff
@@ -12,19 +12,31 @@
 #include "AM_ControlAlgorithm.hpp"
 #include "CommonDataTypes.hpp"
 
+#include "FreeRTOS.h"
```
Aidan-B, line 15: Does `"FreeRTOS.h"` this need to be included in this file?

Aidan-B (changes requested):

AttitudeManager/Inc/AM.hpp
lekangwang: Yes, in `semphr.h`:
Screenshot 2023-07-24 at 9 05 01 PM

AttitudeManager/Src/AM.cpp
```diff
@@ -3,12 +3,37 @@
...
+    .roll = 0.0f,
+    .pitch = 0.0f,
+    .yaw = 0.0f,
+    .throttle = 0.0f
+};
+
+void AttitudeManager::setControlInputs(const AttitudeManagerInput& new_control_inputs) {
+    xSemaphoreTake( control_inputs_mutex, (TickType_t) 10);
+    control_inputs = new_control_inputs;
+    xSemaphoreGive( control_inputs_mutex );
```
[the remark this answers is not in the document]
lekangwang: I just made the mutexes wait indefinitely with a delay of `portMAX_DELAY`

AttitudeManager/Src/AM.cpp
```diff
@@ -20,16 +20,18 @@ AttitudeManagerInput AttitudeManager::control_inputs = {
...
-    control_inputs = new_control_inputs;
-    xSemaphoreGive( control_inputs_mutex );
+    if (xSemaphoreTake(control_inputs_mutex, (TickType_t) portMAX_DELAY) == pdPASS) {
+        control_inputs = new_control_inputs;
+        xSemaphoreGive(control_inputs_mutex);
+    }
 }
 
 AttitudeManagerInput AttitudeManager::getControlInputs() {
     struct AttitudeManagerInput temp;
```
Aidan-B, line 30: Since we return this if we fail to get the mutex, we should probably explicitly default initialize it. I'm also not sure we need to forward declare it.
```suggestion
    AttitudeManagerInput temp{};
```

Aidan-B (changes requested):

AttitudeManager/Src/AM.cpp
lekangwang: having temp is so we can run `xSemaphoreGive(control_inputs_mutex); ` before returning

Chrisc110 (approved):
Just took a look, looking really good!

Aidan-B (approved):

PR #17, opened 2023-07-04
Model configuration system

antimonyB:
## Description

Initial setup of configuration files for a placeholder model.
The motor driver configuration could not be implemented as initially planned.
Further implementation of the configuration system is in progress.

<br/>

This provides a working example of how configuration files will be used.

<br/>

## Testing

The new configuration files compile successfully.
Testing is still in progress.

<br/>

Testing is still in progress.

<br/>

## Documentation

*Milestone number and name:*
Milestone 1: Manual Fixed Wing
*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204811987227184
https://uwarg-docs.atlassian.net/wiki/spaces/Meetings/pages/2266234887/2023-06-28+Model+CFG+2
Further documentation is in progress.
<br/>

## Reminders

- [ ] Add reviewers to the PR


Models/global_config.hpp
Aidan-B, line 1: I think it would probably be wise to wrap this all in a `Config` namespace

Models/global_config.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+typedef enum {
+    positive,
+    negative
+} ControlDirection_t;
+
+//TODO: Determine correct location for motor driver definitions, and finish implementation.
+class MotorDriver{};
+class TempPWMDriver : public MotorDriver{};
+class TempDSHOTDriver : public MotorDriver{};
```
Aidan-B, line 18: I think this is ongoing in #14

Models/testmodel1/config.cpp
```diff
@@ -0,0 +1,49 @@
+#include "config.hpp"
+
+//TODO: Consider where these should be initialized
+TempPWMDriver PWMDriver;
+TempDSHOTDriver DSHOTdriver;
```
Aidan-B, line 5: I think the discussion suggested that we initialize these in the module that uses them (Attitude Manager)

Models/global_config.hpp
```diff
@@ -0,0 +1,51 @@
...
+typedef enum {
+    yaw,
+    pitch,
+    roll,
+    thrust
+} ControlAxis_t;
+
+typedef enum {
+    positive,
+    negative
```
Aidan-B, line 12: Would this be worthwhile? (so you can just multiply by positive or negative to determine motor direction)
```suggestion
    positive = 1,
    negative = -1
```

Alternatively, direction could be replaced with a boolean `inverted` where it is used.

Models/global_config.hpp
```diff
@@ -0,0 +1,51 @@
...
+class TempPWMDriver : public MotorDriver{};
+class TempDSHOTDriver : public MotorDriver{};
+
+typedef struct {
+    ControlAxis_t axis;
+    ControlDirection_t direction;
+    MotorDriver driver;
+    //TODO: Consider a more specific datatype for interfaceID
+    //(e.g. a datatype defined within the motor driver class)
+    uint8_t interfaceID;
```
Aidan-B, line 26: I assume this will ultimately become the constructor arguments for the driver. (e.g. for pwm will be something like timer handle & timer channel). I don't mind the idea of creating this datatype defined in the motor driver class, but we should consider that it could be different for each driver, and so we need a way to vary it based on the driver we are going to use. 

Whatever decision we make here will likely also propagate out to the sensor drivers as well, so it makes sense that we are generic about how we handle this if possible.

Models/testmodel1/config.cpp
```diff
@@ -0,0 +1,49 @@
...
+    },
+    {   //Roll BLDC motor
+        .axis = roll,
+        .direction = negative,
+        .driver = DSHOTdriver,
+        .interfaceID = 0
+    }
+};
+
+const uint8_t NUM_MOTORS = sizeof(motors)/sizeof(Motor_t);
```
Aidan-B, line 22: is it possible to mark this `constexpr` instead?

Tools/Firmware/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 set(ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 
+## TODO: Make this configurable via build script
+Set(MODEL_NAME, testmodel1)
```
Aidan-B, line 11: Looks like this should be pretty easy by passing in a `-DMODEL_NAME` value in the build script
https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-D

Models/testmodel1/config.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+PID_t flightmode1Tuning = {
+    .p = 1.0f,
+    .i = 1.0f,
+    .d = 1.0f
+};
+
+ControlLimits_t flightmode1Limits = {
```
Aidan-B, line 30: I assumed `flightmode1Tuning` and `flightmode1Limits` would be in a struct together in a list of all flight modes, and also including a reference to the flight mode itself. (similar to how you have done for motors, just with a few nested structs)

Models/testmodel1/config.cpp
Aidan-B, line 1: Actually, I think it would be best if we can mark everything in this file `constexpr`, or at the very least `const`

Models/global_config.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+typedef struct {
+    AxisLimit_t yawLimit;
+    AxisLimit_t pitchLimit;
+    AxisLimit_t rollLimit;
+    AxisLimit_t thrustLimit;
+} ControlLimits_t;
+
+//TODO: Combine flightmode classes with AM control algorithms
+class Flightmode{
```
Aidan-B, line 48: This is a class with no public members

Aidan-B: I do understand that it's a placeholder, but It may be more appropriate to just rename the `ControlAlgorithm` class and include it here?

Models/testmodel1/config.cpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+};
+
+const uint8_t NUM_MOTORS = sizeof(motors)/sizeof(Motor_t);
+
+PID_t flightmode1Tuning = {
+    .p = 1.0f,
+    .i = 1.0f,
+    .d = 1.0f
+};
```
Aidan-B, line 28: This unfortunately is unlikely to be sufficient. We will probably need PIDs for each axis that we are attempting to control. I'm not sure how best we handle this.

Maybe each flight mode has an enum of the axes it is controlling, then we define an array with PIDs for each axis in that enum?

Models/testmodel1/config.hpp
```diff
@@ -0,0 +1,8 @@
+#include "global_config.hpp"
+
+extern Motor_t motors[];
+extern const uint8_t NUM_MOTORS;
+
+//TODO: Combine flightmode classes with AM control algorithms
```
Aidan-B, line 6: Looks like there is some stuff that was defined in config.cpp that isn't declared here.

Models/testmodel1/config.hpp
```diff
@@ -0,0 +1,8 @@
+#include "global_config.hpp"
+
+extern Motor_t motors[];
+extern const uint8_t NUM_MOTORS;
```
Aidan-B, line 4: Why are we using `extern` as opposed to just declaration here and definition in the cpp file?

Aidan-B (commented):
Good start. config.cpp looks surprisingly easy to read. Hopefully when it's fully fleshed out it should be pretty straightforward

Models/global_config.hpp
antimonyB: I went with the boolean.

Models/testmodel1/config.cpp
antimonyB: Done, and this allowed everything to live in the header so the cpp file is no longer needed.

Models/testmodel1/config.cpp
antimonyB: I expanded it now to apply to each axis in the same way the control limits do.

Models/testmodel1/config.hpp
antimonyB: I think this was needed to make the globals accessable from other files, but with constexpr its no longer neccessary.

Models/testmodel1/config.cpp
Aidan-B: I think it's possible that PIDs will be needed for axis that aren't just the physical orientation (e.g. speed, altitude?), so I'm not quite sure if that works either. It might be good enough for now though.

Models/global_config.hpp
Aidan-B, line 1: Could we maybe specify default values for each of these structs? that way a model config doesn't need to specify something that it a) doesn't use or b) would be common for most aircraft

antimonyB: It's possible but would make the structs non-aggregate, which makes designated initializers (.fieldName = value) not allowed. And I think those help a lot with the readability of config.hpp.

antimonyB: Actually nevermind, with C++20 we can do both. Added default values.

Models/global_config.hpp
antimonyB: I defined an interface identifier (IID) for each specific driver that inherits a more generic IID depending on the general type of driver. These definitions would have to be moved to the actual drivers. Outside of config and specific driver classes, the IID's can be referred to using their generic base class, so that knowledge of the specific driver type isn't needed.

Models/global_config.hpp
antimonyB: Yes I plan to something along those lines when integrating with AM.

Models/testmodel1/config.cpp
antimonyB: Does the flight mode organization look better now?

Models/global_config.hpp
Aidan-B: So I assume the IID would include information that such as the peripheral handle (e.g. `htim1`) and the channel that is used (e.g. `TIM_CHANNEL_2`) and the structure of the IID would be defined by the driver?

Models/global_config.hpp
```diff
@@ -13,63 +13,127 @@ namespace config
...
+        AxisPID_t pitchPID = {};
+        AxisPID_t rollPID = {};
+        AxisPID_t thrustPID = {};
     } ControlPID_t;
 
     typedef struct {
-        float min;
-        float max;
+        float min = 0.0f;
+        float max = 100.0f;
```
Aidan-B, line 112: I think [-90, 90] might be more reasonable default limits, assuming each value is in degrees

Models/testmodel1/config.cpp
Aidan-B: I like it!

Aidan-B: I like it, but I'm curious if we could just use the same technique as the drivers use to that AM is the one that owns the instantiated driver, instead of making it static from the config file?
If not, I'm happy with this

Models/global_config.hpp
antimonyB: I intended for these to be percentages, which could be interpreted as needed depending on the flight mode.

Models/global_config.hpp
```diff
@@ -2,61 +2,30 @@
 #define ZPSW3_GLOBAL_CONFIG_HPP
 
 #include <cstdint>
+#include <functional>
```
Aidan-B, line 5: I'm pretty sure this is only required if you use std::function

Models/testmodel1/config.hpp
Aidan-B, line 1: I think putting the flight modes in a format similar to how we have it for the drivers would be easier to work with.

Models/global_config.hpp
Aidan-B: I'm not sure how percentages would be interpreted by a flight mode... 100% of 360deg? I'm fine with this if we can communicate that it is a percentage. I wouldn't be opposed to adding a percentage type maybe?

Models/global_config.hpp
```diff
@@ -70,29 +39,53 @@ namespace config
...
     //Define factory function to create motor drivers.
     //The driver class must be copy-constructable as the returned object will be a copy.
-    template <class BaseDriverType, class DerivedDriverType>
-    BaseDriverType constructDriver() {
-        DerivedDriverType driver;
-        return driver;
+    template <class BaseDriverType, class DerivedDriverType, auto... args>
+    BaseDriverType* constructDriver() {
+        return new DerivedDriverType(args...);
     }
```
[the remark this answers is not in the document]
antimonyB: I tried that and got an error due to the driverConstructor field in the Motor_t struct not matching. The struct needs to use the general motor type because it doesn't know what specific type of motor constructor it will hold.

Models/global_config.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+
+
+    /* Flightmode declarations */
+
+    //Define factory function to create flightmodes
+    template <class FlightmodeType, auto... args>
+    AM::Flightmode* constructFlightmode() {
+        return new FlightmodeType(args...);
+    }
```
Aidan-B, line 85: This seems functionally equivalent to `constructDriver`?

Aidan-B (changes requested):
This is looking awesome. Perhaps we can pull the temporary configuration definitions out into their own header that we can slowly remove as more functionality is added instead of constantly modifying the global_config file.

Models/global_config.hpp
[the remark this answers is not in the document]
Aidan-B: I wonder if this is just a difference between how `std::function` and  the `type (*)(args)` format works. I'm fine to keep it either way.

Models/global_config.hpp
antimonyB: Would it be better then to just use the same function template as the drivers, passing Flightmode and ExampleFlightmode1 as BaseDriverType and DerivedDriverType?

Aidan-B: I think so, and we can just use a more generic name for the function that applies to all scenarios.

AttitudeManager/FlightModes/Inc/AM_ControlAlgorithm.hpp
```diff
@@ -13,9 +13,9 @@
```
SnackkOverflowError, line 13: file name needs to be lowercase I believe: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards

Models/testmodel1/config.hpp
```diff
@@ -0,0 +1,133 @@
...
+        }
+    };
+    
+    constexpr uint8_t NUM_GPS = sizeof(GPSArray)/sizeof(GPS_t);
+
+
+
+    /* Flightmode Config */
+
+    class TempFlightmode1 : public AM::Flightmode{
```
SnackkOverflowError, line 64: Are these meant to be defined in here or in the AttitudeManager/FlightModes/Inc?

Models/global_config.hpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    using percentage_t = float;
+
+        class TempDSHOTDriver : public MotorChannel
+        {
+            public:
+                TempDSHOTDriver() {}
+                void set(uint8_t percent){}
+        };
+    class RCInputDriver{};
```
SnackkOverflowError, line 19: Is this necessary? Seems like a large chunk of this code is examples that should be either documented better with explanations or part of a confluence doc or something.

SnackkOverflowError (commented):
Feel like a lot of this needs either better comments or some proper documentation. A lot of temp stuff is just kinda in there which adds to my confusion, and makes the whole thing seem less polished I guess? As someone who doesnt use templates and factories very often, Im not really sure whats going on. Also would be nice if you put labels or wrote in the docs what files are the actual configs that are meant to be read and updated by people (these files should look very simple and clean) and what files support those configs with all the typdefs, templates, and factories.

Chrisc110 (approved):

Models/testmodel1/config.hpp
antimonyB: The idea behind having them here was that any reference to a flight mode that isn't compatible would be caught during compilation.

AttitudeManager/FlightModes/Inc/AM_ControlAlgorithm.hpp
antimonyB: Should this be done as part of this PR? As well as AM.hpp -> am.hpp?

Models/global_config.hpp
antimonyB: I reorganized it so the example code is isolated and can be removed when an actual model is set up. Keeping an example configuration so there is something to compile for now.

Models/example_model/config.hpp
```diff
@@ -0,0 +1,132 @@
...
+        void run();
+        void updatePid();
+    };
+
+    class ExampleFlightmode2 : public AM::Flightmode{
+        public:
+        ExampleFlightmode2(){}
+        void run();
+        void updatePid();
+    };
```
Aidan-B, line 77: Can we move these to the temp_drivers.hpp file and add a todo to remove the temp_drivers file when everything is implemented? Otherwise LGTM

Models/config_foundation.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+
+
+    /* Motor declarations */
+
+    typedef enum {
+        yaw,
+        pitch,
+        roll,
+        thrust
```
Aidan-B, line 31:
```suggestion
        throttle,
        num_controlaxis
```

Models/config_foundation.hpp
```diff
@@ -0,0 +1,99 @@
...
+        percentage_t min = 0.0f;
+        percentage_t max = 100.0f;
+    } AxisLimits_t;
+
+    typedef struct {
+        AxisLimits_t yawLimit = {};
+        AxisLimits_t pitchLimit = {};
+        AxisLimits_t rollLimit = {};
+        AxisLimits_t thrustLimit = {};
+    } ControlLimits_t;
```
antholuo, line 85: Are there going to be structs for ratelimits? ie max rate of rotation.

Is it necessary / was there a discussion?

Aidan-B: It was not discussed. I think it can be added when it's needed (when we create a flight mode that uses rates)

SnackkOverflowError (approved):

PR #21, opened 2023-07-10
Features/driver/iwdg

PratyushMakkar:
@Aidan-B 

## Description
The pull request is for the independent watchdog driver. The driver can be initialized with either a prescaler/reload value or a IDWG watchog typedef, the driver will then copy the values from the watchdog instance The driver allows changing the prescaler/reload values while the program is running and to refresh the watchdog timer.  
<br/>

The driver is part of Milestone 1
<br/>

## Testing
Since the driver is not accessible in debug mode, to test it, I set an indicator LED to turn on and off at the start of the program so indicate that the CPU restarted. One test was to see if prescaler values can be modified while the program ran. To do that, the IOC file was configured to set the prescaler as 16 and reload value as 4095, in the program, the prescaler was set to 8 and reload as 2095. It was then manually checked that the CPU restarted more often (almost instantly) when the prescaler value was lower.

The second test was to refresh the watchdog a fixed amount of times with a HAL_Delay(); and then not to refresh the watchdog. The expected result is that the CPU will restart after 5 seconds + The Independent watchdog timer delay. 

```cpp
HAL_GPIO_TogglePin(GPIOA, GPIO_PIN_5);
   HAL_Delay(500);
   HAL_GPIO_TogglePin(GPIOA, GPIO_PIN_5);
   IWDG_t iwdg{IWDG_PRESCALER_8, 2095};
   iwdg.InitWatchdog(&hiwdg);
   while (1) {}
```
**Second Test**

```cpp
HAL_GPIO_TogglePin(GPIOA, GPIO_PIN_5);
   HAL_Delay(500);
   HAL_GPIO_TogglePin(GPIOA, GPIO_PIN_5);
   while (1) {
         for (uint8_t i = 0; i<100; i++) {
               if (i<10) {
                     IWDG_t::RefreshWatchdog(&hiwdg);
                }
                HAL_Delay(500);
         }
  }
```
<br/>

## Documentation

*Milestone number and name:*
Milestone 1, Name: RSSI/Watchdog 
*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204645025850461
TODO: Unsure how to use confluence. 
<br/>

## Reminders

- [ ] Add reviewers to the PR

### DIFF
--- a/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
+++ b/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
@@ -1,0 +1,27 @@
+/*
+ * IndependentWatchdog.h
+ *
+ *  Created on: Jun. 22, 2023
+ *      Author: pratyushmakkar
+ */
+
+#ifndef SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+#define SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+
+#include <stdint.h>
+#include "../../boardfiles/discoveryl562qe/Drivers/STM32L5xx_HAL_Driver/Inc/stm32l5xx_hal_iwdg.h"
+
+class IWDG_t {
+	private:
+		uint8_t iwdg_prescaler;
+		uint32_t iwdg_reload;
+
+	public:
+		IWDG_t(IWDG_HandleTypeDef* watchdog);
+		IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload);
+		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog, IWDG_t* cfg);
+		static HAL_StatusTypeDef RefreshWatchdog(IWDG_HandleTypeDef* watchdog);
+};
+
+
+#endif /* SRC_DRIVERS_INDEPENDENTWATCHDOG_H_ */
--- a/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
+++ b/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
@@ -8,20 +8,22 @@
 #ifndef SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
 #define SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
 
+
 #include <stdint.h>
-#include "../../boardfiles/discoveryl562qe/Drivers/STM32L5xx_HAL_Driver/Inc/stm32l5xx_hal_iwdg.h"
+#include "main.h"
 
 class IWDG_t {
 	private:
 		uint8_t iwdg_prescaler;
-		uint32_t iwdg_reload;
+		uint16_t iwdg_reload;
 
 	public:
 		IWDG_t(IWDG_HandleTypeDef* watchdog);
 		IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload);
-		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog, IWDG_t* cfg);
+		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog);
 		static HAL_StatusTypeDef RefreshWatchdog(IWDG_HandleTypeDef* watchdog);
 };
 
 
 #endif /* SRC_DRIVERS_INDEPENDENTWATCHDOG_H_ */
+
--- a/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
+++ b/Drivers/IWDG_driver/Inc/IndependentWatchdog.h
@@ -19,7 +19,7 @@ class IWDG_t {
 
 	public:
 		IWDG_t(IWDG_HandleTypeDef* watchdog);
-		IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload);
+		IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload);
 		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog);
 		static HAL_StatusTypeDef RefreshWatchdog(IWDG_HandleTypeDef* watchdog);
 };
--- a/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
+++ b/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
@@ -16,7 +16,7 @@ IWDG_t::IWDG_t(IWDG_HandleTypeDef *watchdog) {
  * @brief Initalize the driver using custom IWDG_Prescaler and realod values.
  *
  */
-IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload) {
+IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload) {
   // assert(IS_IWDG_PRESCALER(iwdg_prescaler)) Should this be added since it is needed when reinitialzing the watchdog
   // assert(IS_IWDG_RELOAD(iwdg_reload))
 	this->iwdg_prescaler = iwdg_prescaler;
--- a/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
+++ b/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
@@ -1,0 +1,45 @@
+#include "../Inc/IndependentWatchdog.h";
+
+
+/**
+ * @brief Borrow the configuration of an already existing watchdog and set it 
+ * to the prescaler and Reload values.
+ * 
+ */
+IWDG_t::IWDG_t(IWDG_HandleTypeDef *watchdog) {
+	IWDG_InitTypeDef *initDef = watchdog->Init;
+	this->iwdg_prescaler = initDef->Prescaler;
+	this->iwdg_reload = initDef->Reload;
+}
+
+/**
+ * @brief Initalize the driver using custom IWDG_Prescaler and realod values. 
+ * 
+ */
+IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload) {
+  // assert(IS_IWDG_PRESCALER(iwdg_prescaler)) Should this be added since it is needed when reinitialzing the watchdog
+  // assert(IS_IWDG_RELOAD(iwdg_reload))  
+	this->iwdg_prescaler = iwdg_prescaler;
+	this->iwdg_reload = iwdg_reload;
+}
+
+/**
+ * @brief Reset the reload and counter values inside a watchdog
+ * using the member variables of IWDG_t
+ * 
+ */
+HAL_StatusTypeDef IWDG_t::InitWatchdog(IWDG_HandleTypeDef *watchdog) {
+	watchdog->Init.Reload = this->iwdg_reload;
+	watchdog->Init.Prescaler = this->iwdg_prescaler;
+	HAL_StatusTypeDef refresh = HAL_IWDG_Init(watchdog);
+	return refresh;
+}
+
+/**
+ * @brief Refresh the watchdog in the parameter. It will reset the downcounter
+ * to the value within the reload register.
+ * 
+ */
+HAL_StatusTypeDef IWDG_t::RefreshWatchdog(IWDG_HandleTypeDef *watchdog) {
+	return HAL_IWDG_Refresh(watchdog);
+}
--- a/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
+++ b/Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
@@ -1,24 +1,24 @@
-#include "../Inc/IndependentWatchdog.h";
-
+#include "IndependentWatchdog.h";
+#include "main.h"
 
 /**
- * @brief Borrow the configuration of an already existing watchdog and set it 
+ * @brief Borrow the configuration of an already existing watchdog and set it
  * to the prescaler and Reload values.
- * 
+ *
  */
 IWDG_t::IWDG_t(IWDG_HandleTypeDef *watchdog) {
-	IWDG_InitTypeDef *initDef = watchdog->Init;
-	this->iwdg_prescaler = initDef->Prescaler;
-	this->iwdg_reload = initDef->Reload;
+	IWDG_InitTypeDef initDef = watchdog->Init;
+	this->iwdg_prescaler = initDef.Prescaler;
+	this->iwdg_reload = initDef.Reload;
 }
 
 /**
- * @brief Initalize the driver using custom IWDG_Prescaler and realod values. 
- * 
+ * @brief Initalize the driver using custom IWDG_Prescaler and realod values.
+ *
  */
 IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload) {
   // assert(IS_IWDG_PRESCALER(iwdg_prescaler)) Should this be added since it is needed when reinitialzing the watchdog
-  // assert(IS_IWDG_RELOAD(iwdg_reload))  
+  // assert(IS_IWDG_RELOAD(iwdg_reload))
 	this->iwdg_prescaler = iwdg_prescaler;
 	this->iwdg_reload = iwdg_reload;
 }
@@ -26,7 +26,7 @@ IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint32_t iwdg_reload) {
 /**
  * @brief Reset the reload and counter values inside a watchdog
  * using the member variables of IWDG_t
- * 
+ *
  */
 HAL_StatusTypeDef IWDG_t::InitWatchdog(IWDG_HandleTypeDef *watchdog) {
 	watchdog->Init.Reload = this->iwdg_reload;
@@ -38,7 +38,7 @@ HAL_StatusTypeDef IWDG_t::InitWatchdog(IWDG_HandleTypeDef *watchdog) {
 /**
  * @brief Refresh the watchdog in the parameter. It will reset the downcounter
  * to the value within the reload register.
- * 
+ *
  */
 HAL_StatusTypeDef IWDG_t::RefreshWatchdog(IWDG_HandleTypeDef *watchdog) {
 	return HAL_IWDG_Refresh(watchdog);


Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+	this->iwdg_prescaler = iwdg_prescaler;
+	this->iwdg_reload = iwdg_reload;
+}
+
+/**
+ * @brief Reset the reload and counter values inside a watchdog
+ * using the member variables of IWDG_t
+ *
+ */
+HAL_StatusTypeDef IWDG_t::InitWatchdog(IWDG_HandleTypeDef *watchdog) {
```
SnackkOverflowError, line 31: Is it possible to make this function return a bool? Want to abstract away the HAL stuff.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+	IWDG_InitTypeDef initDef = watchdog->Init;
+	this->iwdg_prescaler = initDef.Prescaler;
+	this->iwdg_reload = initDef.Reload;
+}
+
+/**
+ * @brief Initalize the driver using custom IWDG_Prescaler and realod values.
+ *
+ */
+IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload) {
```
SnackkOverflowError, line 19: Any reason the Init function and this constructor cant be the same? Like, just do the Init function's logic in this constructor rather than have two separate functions?

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+	HAL_StatusTypeDef refresh = HAL_IWDG_Init(watchdog);
+	return refresh;
+}
+
+/**
+ * @brief Refresh the watchdog in the parameter. It will reset the downcounter
+ * to the value within the reload register.
+ *
+ */
+HAL_StatusTypeDef IWDG_t::RefreshWatchdog(IWDG_HandleTypeDef *watchdog) {
```
SnackkOverflowError, line 43: return a bool rather than HAL_StatusTypeDef here. perhaps return false for an error or something like that? Or make a status bool stored in the driver that can be checked by a manager. @TheSpaceDragon @Chrisc110 @Ayounggie  Thoughts?

SnackkOverflowError (changes requested):

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
TheSpaceDragon: Yeah, I agree with changing the return type to bool. It would be nice to have something like didWatchdogRefresh confirmation after the function returns.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
TheSpaceDragon, line 31: Is there a specific use case where having a separate function to re-initialize the watchdog apart from the constructor would be useful?

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+	return HAL_IWDG_Refresh(watchdog);
```
Aidan-B, line 44: Why don't we store the `IWDG_HanldeTypeDef watchdog` as a class member? doing so would mean that we don't need to pass it as a parameter to each of the functions (and the caller won't need to have a reference to both this driver and the IWDG handle.)

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -0,0 +1,29 @@
...
+class IWDG_t {
+	private:
+		uint8_t iwdg_prescaler;
+		uint16_t iwdg_reload;
+
+	public:
+		IWDG_t(IWDG_HandleTypeDef* watchdog);
+		IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload);
+		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog);
+		static HAL_StatusTypeDef RefreshWatchdog(IWDG_HandleTypeDef* watchdog);
```
Aidan-B, line 24: Why is this static?

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -0,0 +1,29 @@
...
+#include "main.h"
+
+class IWDG_t {
+	private:
+		uint8_t iwdg_prescaler;
+		uint16_t iwdg_reload;
+
+	public:
+		IWDG_t(IWDG_HandleTypeDef* watchdog);
+		IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload);
```
Aidan-B, line 22: What's the purpose of having the two different constructors?

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -0,0 +1,29 @@
...
+#define SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+
+
+#include <stdint.h>
+#include "main.h"
+
+class IWDG_t {
+	private:
+		uint8_t iwdg_prescaler;
+		uint16_t iwdg_reload;
```
Aidan-B, line 18: I think we would be better off storing a `IWDG_HandleTypeDef` instead of these members.

Aidan-B (changes requested):
Overall, it seems like there is a little bit of redundancy in the code that should be pretty easy to simplify.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: Hi, I modified it to return  a bool, (refresh==HAL_OK)

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: I have modified the driver so that setting prescaler and reload is the same as re-initializing the watchdog.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: Should I use try catch to supress errors and return false as well in the newest commit?

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: Modifying the watchdog window while the program is running is possible so I thought I might add it just in case.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
PratyushMakkar: I have handled this to only be a single constructor, where the prescaler and reload can be modified later.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
PratyushMakkar: I have modified it to IWDG_HandleTypeDef* in the latest commit.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
PratyushMakkar: After reformatting the driver, the function is no longer static since it uses a member variable, IWDG_HandleTypeDef*

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: I have modified the driver to use IWDG_HanldeTypeDef watchdog  as a member variable.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#ifndef SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+#define SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+
+
+#include <stdint.h>
+#include "main.h"
+
+class IWDG_t {
```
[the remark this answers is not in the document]
PratyushMakkar: Would IWDGDriver be more suitable?

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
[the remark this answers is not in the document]
PratyushMakkar: is clang-format a bash command? I was unable to run it on my system. I am editing on Vscode and I am using tabs.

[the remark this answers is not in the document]
Aidan-B: I'm pretty sure if you run `Ctrl + Shift + I` in vscode it will format the code for you, using our clang-format rules.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -7,39 +7,54 @@
...
-	watchdog->Init.Reload = this->iwdg_reload;
-	watchdog->Init.Prescaler = this->iwdg_prescaler;
-	HAL_StatusTypeDef refresh = HAL_IWDG_Init(watchdog);
-	return refresh;
+
+bool IWDG_t::RefreshWatchdog() {
+	if (!this->watchdog) {
+		return false;
+	}
+	return (HAL_IWDG_Refresh(watchdog) == HAL_OK);
```
Aidan-B, line 43: Do we need to be concerned about thread safety here?
Will multiple threads need to pet the watchdog, and if so is this operation thread safe?

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -1,29 +1,22 @@
...
 		uint16_t iwdg_reload;
 
 	public:
 		IWDG_t(IWDG_HandleTypeDef* watchdog);
-		IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload);
-		HAL_StatusTypeDef InitWatchdog(IWDG_HandleTypeDef* watchdog);
-		static HAL_StatusTypeDef RefreshWatchdog(IWDG_HandleTypeDef* watchdog);
+		void SetIWDGWatchdog(IWDG_HandleTypeDef* watchdog);
+		IWDG_HandleTypeDef* GetIWDGWatchdog(IWDG_HandleTypeDef* watchdog);
+		bool SetWatchdogParameters(uint8_t prescaler, uint16_t reload);
```
Aidan-B, line 17: I think this should be something like `UpdateWatchdogParameters`, since it updates them. Currently it implies that it sets values that haven't been otherwise set.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -7,39 +7,54 @@
...
-IWDG_t::IWDG_t(uint8_t iwdg_prescaler, uint16_t iwdg_reload) {
-  // assert(IS_IWDG_PRESCALER(iwdg_prescaler)) Should this be added since it is needed when reinitialzing the watchdog
-  // assert(IS_IWDG_RELOAD(iwdg_reload))
-	this->iwdg_prescaler = iwdg_prescaler;
-	this->iwdg_reload = iwdg_reload;
+bool IWDG_t::SetWatchdogParameters(uint8_t prescaler, uint16_t reload) {
+	this->iwdg_prescaler = prescaler;
+	this->iwdg_reload = reload;
+
+	if (!IS_IWDG_PRESCALER(prescaler) || !IS_IWDG_RELOAD(reload)) return false;
```
Aidan-B, line 26: Where are these macros defined? I didn't see them in my skim of the HAL docs

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -7,39 +7,54 @@
...
+	watchdog->Init.Reload = reload;
```
Aidan-B, line 27: You should probably check that the reload value is less than the max value for the reload (`0x0FFF` iirc)

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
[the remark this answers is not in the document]
Aidan-B: I like the name `IndependentWatchdog`, like you have named the file. `IndependentWDG` or `IWDG` might also be suitable.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -7,39 +7,54 @@
...
 	this->iwdg_prescaler = initDef.Prescaler;
 	this->iwdg_reload = initDef.Reload;
 }
 
+
 /**
- * @brief Initalize the driver using custom IWDG_Prescaler and realod values.
+ * @brief Reset the reload and counter values inside a watchdog
+ * using the member variables of IWDG_t
  *
```
Aidan-B, line 20:
```suggestion
 * @returns true on success, false on failure
```

Aidan-B (changes requested):
I would suggest thinking about how you intend for users to interact with the watchdog. In my opinion, the primary actions would be to:
a) setup the watchdog. This can be done entirely from the constructor
b) pet the watchdog. This may need to be done from different threads

Think about then how your driver's API can meet the needs of the user.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
Aidan-B: Okay, after further review, I think I understand why you created this function - to allow the user to set the parameters of the watchdog (before starting the watchdog). 
I would argue (and have been basing my reviews off this assumption) that this should configuration should be done by the user before they pass the `IWDG_HandleTypeDef` to the driver, with the Init function being called from the constructor. I believe it to be the general approach that the team has been taking with other drivers thus far. Curious what you and @SnackkOverflowError think about this approach.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
```diff
@@ -7,39 +7,54 @@
...
+	this->watchdog = watchdog;
 }
+
+/**
+ * @brief Getter function for the watchdog
+ *
+ */
+IWDG_HandleTypeDef* IWDG_t::GetIWDGWatchdog(IWDG_HandleTypeDef* watchdog) {
+	return this->watchdog;
+}
```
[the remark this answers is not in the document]
SnackkOverflowError: I agree, setting and getting the watchdog handle arent really important for this driver imo.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
SnackkOverflowError: Hmmm theres a lot of ways to handle this. On one hand, the watchdog parameters can be set in our IOC (which will generate code to configure them in the watchdog handle I believe). Then the handle is passed into this driver and the parameters can be retrieved from the handle. In which case I agree that this function should be named `UpdateWatchDogParameters` .
What exactly is the usecase for resetting the watchdog parameters after the driver has been created though? Is there a scenario where we want to change the frequency that we pet it at? Seems like we should agree on the watchdog parameters once and then never touch them again? If the usecase is being able to configure them in model config, IMO these params should be parameters in the constructor and set there once, then never reset during the programs lifetime.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -0,0 +1,22 @@
+#ifndef SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+#define SRC_DRIVERS_INDEPENDENTWATCHDOG_H_
+
```
SnackkOverflowError, line 3: Please fix the filenames according to our style guide: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards

SnackkOverflowError (changes requested):

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
TheSpaceDragon: I agree, I feel like updating watchdog parameters after initialization could cause some unforeseen issues later on. Though, it doesn't hurt to have the function _when/if_ need arises for that.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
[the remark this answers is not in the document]
TheSpaceDragon: Also, for this function, wouldn't we be passing a watchdog pointer and then returning the same watchdog?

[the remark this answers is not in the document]
PratyushMakkar: This function has been removed

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
```diff
@@ -1,29 +1,22 @@
...
 
-
 #include <stdint.h>
 #include "main.h"
 
 class IWDG_t {
 	private:
+		IWDG_HandleTypeDef* watchdog;
 		uint8_t iwdg_prescaler;
 		uint16_t iwdg_reload;
```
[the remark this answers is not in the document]
PratyushMakkar: This has been addressed. The values are not stored anymore.

Drivers/IWDG_driver/Inc/IndependentWatchdog.h
PratyushMakkar: This function has been removed as requested.

Drivers/IWDG_driver/Src/IndependentWatchdog.cpp
PratyushMakkar: I believe the HAL docs do not mention it even though it is available. I saw that the f4 did not mention it either even though it is available in the stm32f4xx_iwdg.h file. However the function is removed in the newest commit so that should resolve any potential issues.

Drivers/IWDG_driver/Src/independent_watchdog.cpp
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @brief Borrow the configuration of an already existing watchdog and set it
+ * to the prescaler and Reload values.
+ *
+ */
+IndependentWatchdog::IndependentWatchdog(IWDG_HandleTypeDef *watchdog) {
+	this->watchdog = watchdog;
+	IWDG_InitTypeDef initDef = watchdog->Init;
+	this->iwdg_prescaler = initDef.Prescaler;
+	this->iwdg_reload = initDef.Reload;
```
SnackkOverflowError, line 13: iwdg_prescaler and iwdf_reload dont exist anymore (and arent used) so probably remove lines 11,12 and 13 I think

SnackkOverflowError (changes requested):
Looks good except for some code in your contructor

PR #22, opened 2023-07-19
Initialized AM motors and added OutputToMotor method

Penguronik:
## Description

Added motor protocol initialization in AM constructor
Added method for AM to send output to motor through appropriate protocol
Also renamed control_algorithm to control_algorithm_ to better match the style guide

<br/>

To initialize and use the appropriate motor protocol in AM

<br/>

## Testing

IN PROGRESS

<br/>

IN PROGRESS

<br/>

## Documentation

*Milestone number and name:*

*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204799493626724/f

<br/>

## Reminders

- [ ] Add reviewers to the PR

### DIFF
--- a/AttitudeManager/Inc/AM.hpp
+++ b/AttitudeManager/Inc/AM.hpp
@@ -12,6 +12,8 @@
 #include "AM_ControlAlgorithm.hpp"
 #include "CommonDataTypes.hpp"
 
+#include <vector>
+
 namespace AM {
 
 class AttitudeManager {
@@ -26,16 +28,23 @@ class AttitudeManager {
 
    private:
     AttitudeManager();
-    void outputToMotor(uint8_t percent, uint8_t motorIndex);
+    void outputToMotor(config::ControlAxis_t axis, uint8_t percent);
 
     static SemaphoreHandle_t control_inputs_mutex;
 
     static struct AttitudeManagerInput control_inputs;
 
     Flightmode* control_algorithm_;
-    MotorChannel *motorChannels_[config::NUM_MOTORS];
+
+    std::vector<MotorInstance_t> motorInstances_[4]; // 4 is the size of config::ControlAxis_t enum
+    
 };
 
+typedef struct {
+        config::ControlAxis_t axis;
+        bool isInverted;
+        MotorChannel *motorInstance;
+    } MotorInstance_t;
 
 }  // namespace AM
 
--- a/AttitudeManager/Inc/AM.hpp
+++ b/AttitudeManager/Inc/AM.hpp
@@ -33,7 +33,7 @@ class AttitudeManager {
     static struct AttitudeManagerInput control_inputs;
 
     Flightmode* control_algorithm_;
-    MotorChannel *motorChannel_[config::NUM_MOTORS];
+    MotorChannel *motorChannels_[config::NUM_MOTORS];
 };
 
 
--- a/AttitudeManager/Inc/AM.hpp
+++ b/AttitudeManager/Inc/AM.hpp
@@ -12,8 +12,6 @@
 #include "AM_ControlAlgorithm.hpp"
 #include "CommonDataTypes.hpp"
 
-#include <vector>
-
 namespace AM {
 
 class AttitudeManager {
@@ -30,6 +28,11 @@ class AttitudeManager {
     AttitudeManager();
     void outputToMotor(config::ControlAxis_t axis, uint8_t percent);
 
+    typedef struct {
+        MotorChannel *motorInstance;
+        bool isInverted;
+    } MotorInstance_t;
+
     static SemaphoreHandle_t control_inputs_mutex;
 
     static struct AttitudeManagerInput control_inputs;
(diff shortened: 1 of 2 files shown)


AttitudeManager/Src/AM.cpp
```diff
@@ -2,19 +2,34 @@
...
     // Process Instructions
 
     // Run Control Algorithms
-    control_algorithm->run();
+    control_algorithm_->run();
 
     // Write motor outputs
 }
 
+void AttitudeManager::outputToMotor(uint8_t percent, uint8_t motorIndex) {
```
Aidan-B, line 31: I think `motorIndex` isn't a great way to output to a particular motor. Could we consider at a different function to handle this? (probably one that uses some of the other config info that is available)

AttitudeManager/Src/AM.cpp
```diff
@@ -2,19 +2,34 @@
...
 
+
+
 namespace AM {
 
+AttitudeManager::AttitudeManager(Flightmode* control_algorithm):
+    control_algorithm_(control_algorithm)
+{
+    for (uint8_t i{0}; i < config::NUM_MOTORS; i++) {
+        motorChannels_[i] = config::motors[i].driverConstructor();
```
Aidan-B, line 18: I like that we initialize these in the constructor, but I would be curious if we might need to store them in a different structure so that they are easier to address, as per the previous comment.

Aidan-B (changes requested):
Looks like a good start!  I'm interested to see what you can come up with for the `outputToMotor` method

AttitudeManager/Inc/AM.hpp
```diff
@@ -12,19 +12,32 @@
...
     void runControlLoopIteration(const AttitudeManagerInput& instructions);
 
    private:
     AttitudeManager();
+    void outputToMotor(config::ControlAxis_t axis, uint8_t percent);
+
+    Flightmode* control_algorithm_;
 
-    Flightmode* control_algorithm;
+    std::vector<MotorInstance_t> motorInstances_[4]; // 4 is the size of config::ControlAxis_t enum
```
Aidan-B, line 31: You should probably zero-initialize this.

I'm also not convinced that a vector is required here... more below

AttitudeManager/Src/AM.cpp
```diff
@@ -2,19 +2,45 @@
...
+{
+    for (uint8_t i{0}; i < config::NUM_MOTORS; i++) {
+        
+        motorInstances_[config::motors[i].axis].push_back(
+            {
+                .axis = config::motors[i].axis,
+                .isInverted = config::motors[i].isInverted,
+                .motorInstance = config::motors[i].driverConstructor()
+            }
+        );
```
Aidan-B, line 25: I don't think we necessarily need to copy the values from the config to here. What if we instead store the motorInstances in a normal array, and just create a lookup table of the axis that a motor should be on, and the indices of those motors. That way we don't unnecessarily copy data from the config (since we can just read it from the config) while still retaining the ability to set the motor value in a per-axis fashion?

We could determine the number of motors per axis by iterating over the motor list and counting how many are in each axis. I think this would be fine because we would only need to do it once (during initialization) and the maximum number of motors we can have on zeropilot is 16 (e.g. relatively small).

Aidan-B (changes requested):

AttitudeManager/Src/AM.cpp
```diff
@@ -2,19 +2,76 @@
...
     // Run Control Algorithms
-    control_algorithm->run();
+    control_algorithm_->run();
 
     // Write motor outputs
 }
 
+void AttitudeManager::outputToMotor(config::ControlAxis_t axis, uint8_t percent) {
+    for (MotorInstance_t *i{motorReferences_[axis]}; i != motorReferences_[axis + 1]; i++) {
+        i->motorInstance->set(percent);
```
Aidan-B, line 72: I think you might need to do something to deal with the inverted case here

AttitudeManager/Src/AM.cpp
```diff
@@ -2,19 +2,76 @@
...
+                (motorReferences_[config::pitch] + pitchCount)->isInverted = config::motors[i].isInverted;
+                pitchCount++;
+            case config::roll:
+                (motorReferences_[config::roll] + rollCount)->motorInstance = config::motors[i].driverConstructor();
+                (motorReferences_[config::roll] + rollCount)->isInverted = config::motors[i].isInverted;
+                rollCount++;
+            case config::thrust:
+                (motorReferences_[4] + thrustCount)->motorInstance = config::motors[i].driverConstructor();
+                (motorReferences_[4] + thrustCount)->isInverted = config::motors[i].isInverted;
+                thrustCount++;
```
Aidan-B, line 56: Is it possible to make use of the array operator opposed to raw pointer math?
`motorReferences_[config::yaw][yawCount]->motorInstance...`

Aidan-B (changes requested):

AttitudeManager/Inc/AM.hpp
```diff
@@ -12,19 +12,32 @@
...
+    
 };
+
+// !!unsure if the struct definition should be here, would appreciate feedback
+typedef struct {
+        MotorChannel *motorInstance;
+        bool isInverted;
+    } MotorInstance_t;
+
 }  // namespace AM
```
Aidan-B, line 41: I would suggest either placing this as a member struct (and dropping the typedef) of AttitudeManager, or moving it to Common. I doubt it will be used by any other code, so leaving it here is probably best.

PR #25, opened 2023-07-29
Basic System Manager

TheSpaceDragon:
## Description

Basic System Manager is now complete!! This manager now takes input and executes them on motor outputs in an infinite loop. All that is needed to be done is instantiation and the call of the function, `flyManually()`.

<br/>

N/A

<br/>

## Testing

Need help with on-site members with testing :p

<br/>

To be written...
<br/>

## Documentation

Milestone 1: Manual Fixed Wing
Link to Asana task:
https://app.asana.com/0/1203458353737758/1204799650722171

<br/>

## Reminders

- [x] Add reviewers to the PR


.DS_Store
Aidan-B, line 1: Please dont commit environment settings files

Aidan-B: This includes the entirety of the `.metadata` folder

Boardfiles/discoveryl562qe/Core/Src/main.c
```diff
@@ -59,6 +59,7 @@ SD_HandleTypeDef hsd1;
 
 SPI_HandleTypeDef hspi1;
 
+TIM_HandleTypeDef htim2l
```
Aidan-B, line 62: Did cube ide generate this? it wont compile... there's a missing `;`

.DS_Store
Aidan-B: also `Boardfiles/nucleol552zeq/.settings/language.settings.xml`

Boardfiles/nucleol552zeq/Core/Src/usart.c
```diff
@@ -174,16 +174,18 @@ void MX_USART2_UART_Init(void)
...
   huart2.Instance = USART2;
-  huart2.Init.BaudRate = 115200;
-  huart2.Init.WordLength = UART_WORDLENGTH_8B;
-  huart2.Init.StopBits = UART_STOPBITS_1;
-  huart2.Init.Parity = UART_PARITY_NONE;
+  huart2.Init.BaudRate = 100000;
+  huart2.Init.WordLength = UART_WORDLENGTH_9B;
+  huart2.Init.StopBits = UART_STOPBITS_2;
+  huart2.Init.Parity = UART_PARITY_EVEN;
   huart2.Init.Mode = UART_MODE_TX_RX;
```
Aidan-B, line 181: In your PR description, could you please explain what board settings you have changed, and why you have changed them? Looking at this, I dont understand why we would use a non-standard  BaudRate of 100000

Drivers/rc_receiver/Inc/rcreceiver_datatypes.h
```diff
@@ -37,6 +37,24 @@ struct RCControl{
 
     float &operator[] (int i) { return ControlSignals[i]; }
 
+    void assignValues(RCControl const  &original){
```
Aidan-B, line 40: I don't really understand why this function is needed. I would think to either use the copy constructor, or overload the default equality operator and copy the contents of the `ControlSignals` buffer instead of the value of each element individually.

SystemManager/Src/SystemManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+void SystemManager::flyManually() {
+    for(;;){
+            updateRCInputs();
+            executeInputs();
+    }
+}
+
+void SystemManager::updateRCInputs() {
+    this->rcInputs_.assignValues(rcController_.GetRCControl());
```
Aidan-B, line 32: If we use a copy operator here we could just use
```c++
this->rcInputs_ = rcController_.GetRCControl();
```

SystemManager/Src/SystemManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+    watchdog_.refreshWatchdog();
```
Aidan-B, line 33: This watchdog only asserts that the function is called, not that we are still getting values from the RC. Is this the desired behaviour?

SystemManager/Src/SystemManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "independent_watchdog.h"
+#include "tim.h"
+
+SystemManager::SystemManager():
+    rcController_(&huart2),
+    throttleMotorChannel_(&htim2, TIM_CHANNEL_1),
+    yawMotorChannel_(&htim2, TIM_CHANNEL_2),
+    rollMotorChannel_(&htim2, TIM_CHANNEL_3),
+    pitchMotorChannel_(&htim2, TIM_CHANNEL_4),
+    watchdog_(&hiwdg)
```
Aidan-B, line 20: I'm fine with the way this is setup, but it defeats the purpose of the model config system so I'm confused why you need to be blocked on the including the model config as part of this  PR?

SystemManager/Src/SystemManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+{}
+SystemManager::~SystemManager() {}
+
+void SystemManager::flyManually() {
+    for(;;){
+            updateRCInputs();
+            executeInputs();
+    }
```
Aidan-B, line 28: I understand the appeal of this clean code approach, but I'm not really a fan. It tells you what you're doing, but not how. The two functions you abstracted this into are not complex at all and the logic could absolutely be contained in this function.

The reason I say this is because in `updateRcInputs` you update the inputs, but we then have no clue what is done with them once they have been updated, and in `executeInputs` we update the motor channels, but that doesn't tell us where the values we are updating came from. 

It requires  someone who is new to the code to understand it ALL, and as this grows it will become very confusing. All that is to say, I don't think we need to abstract these into functions, and we can abstract them later once we notice there is code that is being re-used in order to simplify it. If you disagree with me, I would be happy to discuss more.

SystemManager/Src/SystemManager.cpp
```diff
@@ -0,0 +1,43 @@
...
+    this->rcInputs_.assignValues(rcController_.GetRCControl());
+    watchdog_.refreshWatchdog();
+    return;
+}
+
+void SystemManager::executeInputs() {
+    this->throttleMotorChannel_.set(rcInputs_.throttle);
+    this->yawMotorChannel_.set(rcInputs_.yaw);
+    this->rollMotorChannel_.set(rcInputs_.roll);
+    this->pitchMotorChannel_.set(rcInputs_.pitch);
```
Aidan-B, line 41: 2 notes
1. This assumes we only have 1 motor per axis, which wont be true for a plane. We will likely have 2 motors for roll.
2. There is no provision here for arm/disarm. The throttle motor will spin up if the throttle stick on the controller is bumped, and is not safe at all. (this is something that happens all the time, please add an arm/disarm check. Typically you can only arm if the throttle is down and a switch is hit.)

Tools/Firmware/CMakeLists.txt
Aidan-B, line 1: I'm having trouble deciphering here what comes from your changes and what comes from the model config pr, but it looks like there was some accidental formatting changes.

Aidan-B (changes requested):

Boardfiles/nucleol552zeq/Core/Src/usart.c
TheSpaceDragon: These settings are from the SBUS documentation outlining the desired configurations. Heres a link to that https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2251456686/SBUS+Receiver

Drivers/rc_receiver/Inc/rcreceiver_datatypes.h
TheSpaceDragon: Yup, that is a good idea. I'll try to implement that 👍

SystemManager/Src/SystemManager.cpp
TheSpaceDragon: I'll revert the model config changes for this one, forgot to undo those from before lol

SystemManager/Src/SystemManager.cpp
TheSpaceDragon: That sounds good, I can just remove the abstraction

.DS_Store
TheSpaceDragon: Just removed these files!

Boardfiles/discoveryl562qe/Core/Src/main.c
TheSpaceDragon: I think that might've been there by accident. Fixed.

SystemManager/Src/SystemManager.cpp
TheSpaceDragon: Updated it so that SM only kicks the watchdog when new data is received

SystemManager/Src/SystemManager.cpp
TheSpaceDragon: 1. I believe for M1 I was asked to assume that each axis has its own motor.  
2. Just added a provision to check for arm/disarm

Tools/Firmware/CMakeLists.txt
TheSpaceDragon: Reverted Model Config changes. It should be gone now

SystemManager/Src/SystemManager.cpp
Chrisc110: @TheSpaceDragon has this been changed?

TheSpaceDragon: The model config changes have been reverted now. These are directly interfacing with the IOC and drivers. I believe the plan is to implement this with model-config for later PRs

Drivers/rc_receiver/Inc/rcreceiver_datatypes.h
```diff
@@ -2,6 +2,7 @@
 #define RC_RECEIVER_DATA_TYPES_H
 
 #include <cstdint>
+#include <iostream>
```
Aidan-B, line 5: I'm surprised this is the header that is required for your change. cppreference.com suggests that `std::copy` is from \<algorithm\>

Drivers/rc_receiver/Src/sbus_receiver.cpp
```diff
@@ -12,6 +12,7 @@ SBUSReceiver::SBUSReceiver(UART_HandleTypeDef* uart) : uart_(uart)
     received_sbus_.failsafe = false;
     received_sbus_.lost_frame = false;
     received_sbus_.new_data = false;
+    this->setIsDataNew(false);
```
Aidan-B, line 15: perhaps you can just use the initializer list to set the value of `is_data_new_`

Drivers/rc_receiver/Inc/sbus_receiver.hpp
```diff
@@ -37,6 +37,8 @@ class SBUSReceiver{
         public variable
     */
         uint8_t raw_sbus_[SBUS_FRAME_SIZE];
+        bool getIsDataNew();
+        void setIsDataNew(bool newVal);
```
Aidan-B, line 41: I don't think this set method is required. Just update the variable in the `parse()` method, and whenever the data is read

SystemManager/Inc/SystemManager.hpp
```diff
@@ -10,6 +10,10 @@
 #include "independent_watchdog.h"
 #include "ZP_D_PWMChannel.hpp"
 
+#define SBUS_THRESHOLD 25
+#define SBUS_MIN 0
+#define SBUS_MAX 100
+
```
Aidan-B, line 16: Why are these defined here instead of the SBUS class?

SystemManager/Src/SystemManager.cpp
```diff
@@ -19,25 +19,26 @@ SystemManager::SystemManager():
...
+            this->yawMotorChannel_.set(rcInputs_.yaw);
+            this->rollMotorChannel_.set(rcInputs_.roll);
+            this->pitchMotorChannel_.set(rcInputs_.pitch);
+        }
+        else if (this->rcInputs_.arm <= SBUS_THRESHOLD) {
+            this->throttleMotorChannel_.set(0);
+            this->yawMotorChannel_.set(0);
+            this->rollMotorChannel_.set(0);
+            this->pitchMotorChannel_.set(0);
+        }
```
Aidan-B, line 41: What happens if the arm channel reads between the thresholds? I think this should just be a straight `if {} else {}` scenario

Aidan-B (changes requested):

SystemManager/Src/SystemManager.cpp
Aidan-B: I can say with near certainty that you will need to add a second motor for the roll axis that is inverted (1 motor per aileron)

Drivers/rc_receiver/Inc/sbus_receiver.hpp
SnackkOverflowError: Agree, the isDataNew bool should not be able to be set by SM or anything besides sbus driver.

SystemManager/Src/SystemManager.cpp
SnackkOverflowError: this was discussed at the meeting actually, @Chrisc110 mentioned doing it this way in case the arm (or any other binary channel) was some value in between 0 and 100. is there a guaruntee that itll be either one or the other?

SystemManager/Inc/SystemManager.hpp
SnackkOverflowError: either these are in SM, or in SBUS and exposed to SM, but it sounds like the only component that needs these values is SM so i see arguments for keeping it here and arguments for putting it in the sbus driver.

SystemManager/Src/SystemManager.cpp
TheSpaceDragon: The values for arm/disarm are binary (either 0 or 100) so we added some threshold in case there is some noise around it. If its in the middle somewhere, we just ignore it I believe

Aidan-B: My argument is that we should do something in the middle. Right now, the drone will lock up and be both uncontrollable and flying at the exact configuration it was before the flaky arm/disarm signal is received.

Aidan-B: For context, someone could bind a 3 position switch to the arm/disarm channel (I know Anthony does this, and hopefully he also adjusts the controller settings to only output 0 or 100) so it wouldn't be impossible to accidentally send 50 on the arm/disarm channel.

antholuo: It's setup so that the arm/disarm will only be 0 or 100, but it is possible for a controller to "flake" and read non-0 or 100 values...

This does not stop other people from doing the same thing and forgetting to do the check and sending 0-50-100.

antholuo: I looked at the code a bit more, I think the safest thing you can do is check for arm if ARM > MAX - thresh. Adding a check to disarm doesn't make sense (let's say you lose RC link and the failsafe isn't configured properly, so you output 50% or 70% or whatever the failsafe is set to).

Drivers/rc_receiver/Src/sbus_receiver.cpp
```diff
@@ -97,4 +98,12 @@ void SBUSReceiver::cast_rccontrol()
...
- }
\ No newline at end of file
+ }
+
+ bool SBUSReceiver::getIsDataNew(){
+    return is_data_new_;
+ }
+
+ void SBUSReceiver::setIsDataNew(bool newVal){
+    this->is_data_new_ = newVal;
```
antholuo: What is the purpose of checking for new data? Just missing context && haven't been following

SystemManager/Inc/SystemManager.hpp
```diff
@@ -0,0 +1,38 @@
...
+private:
+    SBUSReceiver rcController_;
+    RCControl rcInputs_;
+    PWMChannel throttleMotorChannel_;
+    PWMChannel yawMotorChannel_;
+    PWMChannel rollMotorChannel_;
+    PWMChannel invertedRollMotorChannel_;
+    PWMChannel pitchMotorChannel_;
+    IndependentWatchdog watchdog_;
+};
```
antholuo, line 35: Just realized...is there the option to "passthrough" a channel? 

For ex, I just want a PWM output to directly mirror one of the rc inputs, not tied to pitch/roll/yaw?

Drivers/rc_receiver/Src/sbus_receiver.cpp
Aidan-B: Failsafe, in case the receiver gets unplugged

SystemManager/Inc/SystemManager.hpp
Aidan-B: Not currently, but this should probably be a wider discussion which is handled in the model configuration file

SnackkOverflowError (approved):

Chrisc110 (approved):

PR #26, opened 2023-08-07
Unit test sbus

HardyYu:
## Description

*Changes made:*
- sbus unit test added
- fake function framework library added
- cmake changed

*How this unit testing was done*
- tested with basic gtest TEST function
- solved the hal library issue by faking functions from hal by fff

*Things need to be mindful*
- defined UNIT_TEST macro in the unit test cmake. This means in the later unit test, the macro UNIT_TEST can turn on/off the code we what to have separately for unit test
- new library folder is introduced under Common/, not sure if we want to embrace this style

## Documentation

https://github.com/meekrosoft/fff

### DIFF
--- a/Drivers/rc_receiver/Inc/sbus_receiver.hpp
+++ b/Drivers/rc_receiver/Inc/sbus_receiver.hpp
@@ -1,21 +1,9 @@
 #ifndef SBUS_RECEIVER_HPP
 #define SBUS_RECEIVER_HPP
 
+#include "usart.h"
 #include "rcreceiver_datatypes.h"
 #include "sbus_defines.h"
-
-#ifdef UNIT_TEST
-    typedef struct __UART_HandleTypeDef{
-    }UART_HandleTypeDef;
-
-    typedef enum
-    {
-    } HAL_StatusTypeDef;
-    
-    HAL_StatusTypeDef HAL_UART_Receive_DMA(UART_HandleTypeDef *huart, uint8_t *pData, uint16_t Size);
-#else
-    #include "usart.h"
-#endif
 
 /*
  * UART Config for this sbus driver
--- a/Drivers/rc_receiver/Inc/sbus_receiver.hpp
+++ b/Drivers/rc_receiver/Inc/sbus_receiver.hpp
@@ -1,9 +1,21 @@
 #ifndef SBUS_RECEIVER_HPP
 #define SBUS_RECEIVER_HPP
 
-#include "usart.h"
 #include "rcreceiver_datatypes.h"
 #include "sbus_defines.h"
+
+#ifdef UNIT_TEST
+    typedef struct __UART_HandleTypeDef{
+    }UART_HandleTypeDef;
+
+    typedef enum
+    {
+    } HAL_StatusTypeDef;
+    
+    HAL_StatusTypeDef HAL_UART_Receive_DMA(UART_HandleTypeDef *huart, uint8_t *pData, uint16_t Size);
+#else
+    #include "usart.h"
+#endif
 
 /*
  * UART Config for this sbus driver
--- a/Drivers/rc_receiver/Tests/sbus_gtest.cpp
+++ b/Drivers/rc_receiver/Tests/sbus_gtest.cpp
@@ -148,8 +148,8 @@ TEST (RCControlStructTest, NamedArrayElement){
         expectedRC.ControlSignals[7] = 0;
         expectedRC.ControlSignals[8] = 100;
         expectedRC.ControlSignals[9] = 100;
-        expectedRC.ControlSignals[10] = 100.0 / 3.0;
-        expectedRC.ControlSignals[11] = 100.0 / 3.0;
+        expectedRC.ControlSignals[10] = 33.3125;
+        expectedRC.ControlSignals[11] = 33.3125;
         expectedRC.ControlSignals[12] = 25;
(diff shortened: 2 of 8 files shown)


Drivers/rc_receiver/Inc/sbus_receiver.hpp
```diff
@@ -1,10 +1,22 @@
...
+    }UART_HandleTypeDef;
+
+    typedef enum
+    {
+    } HAL_StatusTypeDef;
+    
+    HAL_StatusTypeDef HAL_UART_Receive_DMA(UART_HandleTypeDef *huart, uint8_t *pData, uint16_t Size);
+#else
+    #include "usart.h"
+#endif
```
Chrisc110, line 18: interesting, so the plan is to stub out all the hal functions here? It looks slightly messy especially since its in the actual file... would it be at all possible to have this in a different file? Like a UT_sbus_receiver.hpp file or something? Is that even possible? I'd like to know everyones thoughts...

Chrisc110: After some discussion: having a directory like Testing/Inc/ and Testing/Src/ to put the files related for each stub... for example since you are using a uart function, you would create the uart.h/c file (or whatever the actual HAL file is named) and add your stub function in here. Now people that want to use the same function can reference it from that file, if its not in there, they can append it to the bottom, and if the file doesn't exist they can make it.

Aidan-B: It looks like the fake function framework does something similar to what we discussed. It looks like fff is the real solution here (and it is more robust than just stubbing out a bunch of functions). I think the same concept still holds with fff though (putting the faked functions definitions in their own files, and compiling them in only if we are using a testing build)

Drivers/rc_receiver/Tests/sbus_gtest.cpp
```diff
@@ -0,0 +1,98 @@
...
+    memcpy(sbus_test.raw_sbus_, in_bytes, sizeof(in_bytes));
+    sbus_test.parse();
+
+    for(int i = 0; i < SBUS_INPUT_CHANNELS; i ++){
+        EXPECT_TRUE(sbus_test.GetSBUS().ch[i] == packet.ch[i]);
+    }
+
+    for(int i = 0; i < SBUS_INPUT_CHANNELS; i ++){
+        EXPECT_TRUE(sbus_test.GetRCControl().ControlSignals[i] == 0);
+    }
```
Aidan-B, line 52: I don't quite understand why these are different (Why is GetRCControl always zero?)

I think a code comment may be useful here (or maybe just more test cases would be sufficient)?

Drivers/rc_receiver/Tests/sbus_gtest.cpp
```diff
@@ -0,0 +1,98 @@
...
+FAKE_VALUE_FUNC(HAL_StatusTypeDef, HAL_UART_Receive_DMA, UART_HandleTypeDef*, uint8_t*, uint16_t );
+
+TEST (SBUS_Receiver_Test, RawDataToSBus){
+    const uint8_t in_bytes[SBUS_FRAME_SIZE] = {
+        HEADER_, 0x80, 0x08, 0x84,
+        0x20, 0x06, 0x41, 0x88, 0x42, 0x18,
+        0xE2, 0x10, 0x88, 0x48, 0x84, 0x22,
+        0x16, 0xC1, 0x88, 0x46, 0x38, 0xE2,
+        0x11, 0x00, FOOTER_
+    };
```
Aidan-B, line 20: I think this is a great start, but I think it would be good to see some more test cases. Perhaps thing about what values are edge cases and ensure that the code still works correctly.

Drivers/rc_receiver/Tests/sbus_gtest.cpp
```diff
@@ -0,0 +1,98 @@
...
+        packet.ch[11] = 139,
+        packet.ch[12] = 140,
+        packet.ch[13] = 141,
+        packet.ch[14] = 142,
+        packet.ch[15] = 143
+    };
+
+    UART_HandleTypeDef uart_fake;
+    SBUSReceiver sbus_test(&uart_fake);
+    memcpy(sbus_test.raw_sbus_, in_bytes, sizeof(in_bytes));
```
Aidan-B, line 43: I'm surprised that `raw_sbus_` is public... this should probably be private, and you make use of the fake `HAL_UART_Receive_DMA` to get the address to memcpy into.

Aidan-B (changes requested):
Great stuff! FFF is a great find, and I think it looks very powerful. You're absolutely on the right track with the tests, just add a couple more test cases.

Drivers/rc_receiver/Tests/sbus_gtest.cpp
HardyYu: I made raw_sbus_ to be public because of the nature of DMA callback. The DMA function takes a buffer pointer as a parameter and updates the value when the DMA interrupt was triggered. Also in the callback function, I need to call DMA again to make sure it runs circularly.  Maybe I can change the implementation by adding a getter method that returns the pointer of the buffer, so outside of the scope, the callback function could still modify the content of the buffer.

Chrisc110 (approved):
Approved on basis that all comments are resolved.

PR #30, opened 2023-09-01
Basic AM thread in SM

lekangwang:
## Description

- Add private AM class instance as a SM class member
- Modify `flyManually()` SM member function to set throttle, yaw, pitch and roll via a setter to allow communication with AM which will operate the motors using PWM accordingly (via [outputToMotor()](https://github.com/UWARG/ZeroPilot-SW-3.5/pull/22) method) 
- Added static `runAM()` function to run `runControlLoopIteration()` at set frequency
- Initialize/destroy AM thread in SM constructor/destructor 

<br/>

- Have SM spawn AM thread to use AM and start communicating new RC inputs to AM thread

<br/>

## Testing

- Code compiles
- Did some hardware tests to validate that AM thread runControlLoopIteration(); receives updated values to AM controller struct

<br/>

- None so far 

<br/>

## Documentation

*Milestone number and name:*

*Link to Asana task:*

<br/>

## Reminders

- [x] Add reviewers to the PR


SystemManager/Src/SystemManager.cpp
```diff
@@ -10,37 +10,56 @@
...
     rcController_(&huart2),
     throttleMotorChannel_(&htim2, TIM_CHANNEL_1),
     yawMotorChannel_(&htim2, TIM_CHANNEL_2),
     rollMotorChannel_(&htim2, TIM_CHANNEL_3),
     pitchMotorChannel_(&htim2, TIM_CHANNEL_4),
-    invertedRollMotorChannel_(&htim3, TIM_CHANNEL_1),
-    watchdog_(&hiwdg)
-{}
+    watchdog_(&hiwdg),
+    am_instance_(config::flightmodes[config::DEFAULT_FLIGHTMODE].flightmodeConstructor())
```
Aidan-B, line 24: I think we should probably construct the flight modes all at once in one place, perhaps in the attitude manager constructor?
I actually don't know if we want to have multiple flight modes all allocated at the same time and just move a pointer to the active one or if we want to create / destroy them as needed. I think this is something that needs to be discussed, but the way you have it now is fine until we need to make a decision on this.

SystemManager/Src/SystemManager.cpp
```diff
@@ -10,37 +10,56 @@
...
-            this->pitchMotorChannel_.set(0);
-            this->invertedRollMotorChannel_.set(0);
+        if (this->rcInputs_.arm >= (SBUS_MAX/2)) {
+            am_input.roll = rcInputs_.roll;
+            am_input.pitch = rcInputs_.pitch;
+            am_input.yaw = rcInputs_.yaw;
+            am_input.throttle = rcInputs_.throttle;
+            AM::AttitudeManager::setControlInputs(am_input);
+        } else {
+            AM::AttitudeManager::setControlInputs(am_input);
```
Aidan-B, line 62: Do you think we should add arm and disarm methods directly into AM, so as to ensure that we can't accidentally write some value to the motors if we are in a flight mode that does not just directly map the controller inputs?

SystemManager/Src/SystemManager.cpp
```diff
@@ -10,37 +10,56 @@
...
     yawMotorChannel_(&htim2, TIM_CHANNEL_2),
     rollMotorChannel_(&htim2, TIM_CHANNEL_3),
     pitchMotorChannel_(&htim2, TIM_CHANNEL_4),
-    invertedRollMotorChannel_(&htim3, TIM_CHANNEL_1),
-    watchdog_(&hiwdg)
-{}
+    watchdog_(&hiwdg),
+    am_instance_(config::flightmodes[config::DEFAULT_FLIGHTMODE].flightmodeConstructor())
+{
+    xTaskCreate(runAM, "AM Thread", 400U, (void*)&am_instance_, osPriorityNormal, &AM_handle_);
```
Aidan-B, line 26: I can't remember if anything else needs to be done to start a task (I don't think so though...). Were you able to test this on hardware?

SystemManager/Src/SystemManager.cpp
```diff
@@ -10,37 +10,56 @@
...
+    TickType_t xLastWakeTime;
+    xLastWakeTime = xTaskGetTickCount();
+    for(;;) {
+        am_instance->runControlLoopIteration(AM::AttitudeManager::getControlInputs());
+        vTaskDelayUntil(&xLastWakeTime, AM_PERIOD_MS);
+    }
+}
 
 void SystemManager::flyManually() {
     for(;;){
```
antimonyB, line 44: Where would this be called? I imagine it would ideally run repeatedly at the same rate as runAM so that it has up-to-date control inputs.

SystemManager/Src/SystemManager.cpp
lekangwang: Decided to construct flight mode objects on the fly for now.

SystemManager/Src/SystemManager.cpp
lekangwang: image

This looks promising. 

I'm running the AM thread at a period of 1 second and the set AM instructions thread at a period of 2 seconds (which simply increments the AM control struct using the static getter and setter methods defined in AM).

The AM thread `runControlLoopIteration()` seems to be picking up the new values at the regular interval.

SystemManager/Src/SystemManager.cpp
lekangwang: Are you saying we should add arm/disarm logic to `AM::setControlInputs()`? I'm guessing you want to disarm based on some flightmode property.

SystemManager/Src/SystemManager.cpp
lekangwang: This is called in `main.cpp` when we create a SM thread that calls `flyManually()`

SnackkOverflowError (approved):
lets get this money

HardyYu (approved):

PR #31, opened 2023-09-09
Added OutputToMotor method

Penguronik:
## Description

This PR originally had motor initialization, but that will now be done in SM and added in a separate PR
Added method for AM to send output to motor through appropriate protocol
Also changes to CmakeLists as well as stubs added to make it compilable

<br/>

To initialize and use the appropriate motor protocol in AM

<br/>

## Testing

IN PROGRESS

<br/>

IN PROGRESS

<br/>

## Documentation

*Milestone number and name:*

*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204799493626724/f

<br/>

## Reminders

- [x] Add reviewers to the PR


AttitudeManager/Inc/AM.hpp
```diff
@@ -11,30 +11,48 @@
 
 #include "AM_ControlAlgorithm.hpp"
 #include "CommonDataTypes.hpp"
-
-#include "FreeRTOS.h"
 #include "semphr.h"
+#include "config_foundation.hpp"
+#include "config.hpp"
+#include <gtest/gtest.h>
```
antholuo, line 17: Is there a need to import gtest in deployed files?

AttitudeManager/Src/AM.cpp
```diff
@@ -35,13 +35,93 @@ AttitudeManagerInput AttitudeManager::getControlInputs() {
     return temp;
 }
 
+AttitudeManager::AttitudeManager(Flightmode* control_algorithm):
+    control_algorithm_(control_algorithm),
+    motorInstances_(new MotorInstance_t[config::NUM_MOTORS]{})
+{
+    // Go through motor config list and count the # of motors of each axis type
+    uint8_t yawCount{0}, pitchCount{0}, rollCount{0}, throttleCount{0};
```
antholuo, line 43: I personally think slightly more verbose names might be helpful. Maybe "numYawMotors" or "numMotorsYaw". 

If our style guide is still relevant, I'd take a look at "Member Variables" section here: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1827635203/Style+Guide

AttitudeManager/Inc/AM.hpp
Aidan-B, line 17:
```suggestion
#include <gtest/gtest_prod.h>
```

AttitudeManager/Src/AM.cpp
Aidan-B: I think this may be superseded by https://uwarg-docs.atlassian.net/l/cp/61ADnS0A

Penguronik: Changed those variable to be more verbose, definitely agreed. I originally went with the style guide Aidan sent for my variables but not sure if there is something else you wanted me to improve that is not mentioned there

AttitudeManager/Src/AM.cpp
```diff
@@ -35,13 +35,93 @@ AttitudeManagerInput AttitudeManager::getControlInputs() {
     return temp;
 }
 
+AttitudeManager::AttitudeManager(Flightmode* control_algorithm):
+    control_algorithm_(control_algorithm),
+    motorInstances_(new MotorInstance_t[config::NUM_MOTORS]{})
```
Aidan-B, line 40: As I'm going through this PR, I'm feeling more and more like we should be using dependency injection to pass in the motors (the same way `control_algorithm` is injected. This allows you to inject mock motors in unit tests without needing to friend the class. This design pattern is why we have a parent `MotorChannel` and each driver inherits from it.

AttitudeManager/Tests/AMTest.cpp
```diff
@@ -0,0 +1,295 @@
...
+        EXPECT_THAT(am.motorReferences_[1], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[2], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[3], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[4], ::testing::NotNull());
+
+        // Setting non-existent motors first shouldn't affect roll
+        am.outputToMotor(config::throttle, 40);
+        am.outputToMotor(config::pitch, 30);
+        am.outputToMotor(config::yaw, 20);
+        am.outputToMotor(config::roll, 10);
```
Aidan-B, line 151: You have verified this behaviour already in the previous test. No need to verify it again in this one.

Testing/Stubs/Inc/ZP_D_MotorChannel.hpp
Aidan-B, line 1: You shouldn't need this file. use a mock instead

Testing/Stubs/Inc/ZP_D_PWMChannel.hpp
Aidan-B, line 1: You shouldn't need this file. use a mock instead

Tools/Testing/CMakeLists.txt
```diff
@@ -23,21 +26,24 @@ set(ROOT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../..")
...
 set(AM_FOLDER "${ROOT_DIR}/AttitudeManager")
-set(AM_INC "{$AM_FOLDER}/Inc")
-set(AM_SRC "{$AM_FOLDER}/Src")
+set(AM_INC "${AM_FOLDER}/Inc")
+set(AM_SRC "${AM_FOLDER}/Src")
+set(FLIGHTMODES_INC "${AM_FOLDER}/FlightModes/Inc")
+set(FLIGHTMODES_SRC "${AM_FOLDER}/FlightModes/Src")
 
-include_directories(${AM_INC})
+include_directories(${AM_INC} ${FLIGHTMODES_INC} ${AM_FOLDER}/Tests)
```
Aidan-B, line 34: There aren't any headers in the Tests folder, so no need to include it in the list of include directories
```suggestion
include_directories(${AM_INC} ${FLIGHTMODES_INC})
```

Testing/Stubs/Inc/main.h
Aidan-B, line 1: Using a MotorChannel mock I think you shouldn't need any of this either

Testing/Stubs/Inc/tim.h
Aidan-B, line 1: Shouldn't need this either...

Testing/Stubs/Src/ZP_D_PWMChannel.cpp
```diff
@@ -0,0 +1,12 @@
...
+PWMChannel::PWMChannel(TIM_HandleTypeDef* timer,
+                        uint16_t timerChannel)
+{
+    timerChannel_ = timerChannel;
+    percent_ = 0;
+}
+
+void PWMChannel::set(uint8_t percent) {
+    percent_ = percent;
+}
```
Aidan-B, line 12: Use this and convert it into the mock (and drop the constructor inputs)

Testing/Stubs/Src/main.cpp
Aidan-B, line 1: Shouldn't need this

Testing/Stubs/Inc/FreeRTOS.h
Aidan-B, line 1: I'm not sure this needs to be stubbed either

Aidan-B (changes requested):
Okay, so overall well done. There's a lot of comments, but I think you're still most of the way there. Here are the key takeaways:
1. We can dynamically allocate memory during initialization. It's alright if it's allocated ONLY at initialization, but realistically I don't think you need to allocate the memory because:
2. You should make a mock of MotorChannel. You basically did this, but you called it a stub. Move the class from the stub folder to the appropriate driver folder, change it's name to `MockMotor` or something to that effect, and don't give it any constructor inputs (unless they are needed for some reason, but I can't see why they would be). We spent a bunch of time in designing our drivers around this inheritance/mocking paradigm to make testing as easy as possible.
3. In your unit tests, make sure you are generally testing one thing at a time. So that is to say, if you are testing that giving some input to AM causes an output to be generated that is 1 test. I recognize this means you won't be able to test a few things without other parts of AM being completed (or using Friendship). That's fine. You can add them as TODOs and I can implement them in my PR which will be directly following yours.

AttitudeManager/Inc/AM.hpp
```diff
@@ -11,30 +11,52 @@
 
 #include "AM_ControlAlgorithm.hpp"
 #include "CommonDataTypes.hpp"
-
-#include "FreeRTOS.h"
 #include "semphr.h"
+#include "config_foundation.hpp"
+#include "config.hpp"
+#ifdef TESTING
```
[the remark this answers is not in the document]
Penguronik: I think I defined it [here](https://github.com/UWARG/ZeroPilot-SW-3.5/pull/31/files/362331aea825285f1e043a64c42f408e2efd5910#diff-c270b2f4693b444bd56f97867eaba5b39ea4da0b44d8698356a3ba1c94a95fe1R12) but first time using that command so not sure if I missed anything. Seemed to run tests fine though.

AttitudeManager/Src/AM.cpp
```diff
@@ -4,11 +4,11 @@
 #include "AM.hpp"
 
 #include "FreeRTOS.h"
-#include "semphr.h"
```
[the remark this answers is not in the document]
Penguronik: Accident while clearing some of my own extra headers, added it back

AttitudeManager/Src/AM.cpp
Penguronik: That makes a lot of sense I'll make a commit with that soon

AttitudeManager/Src/AM.cpp
```diff
@@ -35,13 +35,93 @@ AttitudeManagerInput AttitudeManager::getControlInputs() {
...
+                break;
+            case config::roll: 
+                numRollMotors++;
+                break;
+        }
+    }
+
+    // Dedicate portions of the motorInstances array to the respective axis by holding references to those positions
+    // We do this instead of creating 4 seperate arrays as we only know the # of total motors and thus don't know the size of the 4 arrays at compile time
+    // If there is 0 of an axis, having the same reference as the next axis will cause it to do nothing
```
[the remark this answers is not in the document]
Penguronik: Yep definitely think it would be easier to allocate 4 arrays instead now. The only thing would be that now we would be allocating arrays based on a computed value instead of just a constant so I don't know if that would make it more unsafe and less preferable

AttitudeManager/Tests/AMTest.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+        AM::Flightmode *control_algorithm  = config::flightmodes[0].flightmodeConstructor();
+        AM::AttitudeManager am{control_algorithm};
+
+        // Non null references
+        EXPECT_THAT(am.motorReferences_[0], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[1], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[2], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[3], ::testing::NotNull());
+        EXPECT_THAT(am.motorReferences_[4], ::testing::NotNull());
```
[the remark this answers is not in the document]
Penguronik: My thinking was that currently we know an axis has 0 items when the reference to it is the same as the reference to the next axis, if there are no motors then all the references should be an equal value (arr[0] where arr is an array of 0 size, so basically an out of bounds reference) but not null. If we decide to go with using 4 arrays instead that wont be the case though anyways

AttitudeManager/Tests/AMTest.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+        // Shouldn't do anything
+        am.outputToMotor(config::throttle, 40);
+        am.outputToMotor(config::pitch, 30);
+        am.outputToMotor(config::roll, 20);
+        am.outputToMotor(config::yaw, 10);
```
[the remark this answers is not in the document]
Penguronik: I don't think I can as there are no values to edit in the first place, this was more so to ensure it doesn't crash or segfault, not sure if there is a more official way to indicate that that's what I'm testing for

AttitudeManager/Tests/AMTest.cpp
```diff
@@ -0,0 +1,295 @@
...
+    // 5 motors of yaw type, some are inverted, *DO MIXED DRIVERS WHEN THERE ARE IMPLEMENATIONS OF MORE THAN ONE DRIVER*
+    {
+        config::NUM_MOTORS = 5;
+        config::motors = new config::Motor_t[config::NUM_MOTORS];
+
+        config::motors[0] =
+            config::Motor_t{   //Yaw servo motor
+                .axis = config::yaw,
+                .isInverted = false,
+                .driverConstructor = config::constructObject<MotorChannel, PWMChannel,
```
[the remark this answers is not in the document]
Penguronik: My only problem with this is that as the driverConstructor returns a pointer to MotorChannel base class, I'm not able to access any of the variables or methods I created inside my MockMotor class for testing purposes (such as a way to display the value it would normally set in __HAL_TIM_SET_COMPARE)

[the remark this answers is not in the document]
Penguronik: Otherwise it makes sense that this is the best way to do it and I would prefer doing it this way, just not sure how to solve that problem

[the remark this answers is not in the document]
Penguronik: Also placing the mock alongside the other motor types makes it that I have to compile PWMChannel too even though I'm not using it, which gives me compiler errors for missing SystemCoreClock, __HAL_TIM_SET_PRESCALER, etc which I can just stub but feels like I shouldn't need to. Not sure if it would be an acceptable solution to just change cmakelists from "${MOTORCHANNEL_SRC}/*.cpp" to "${MOTORCHANNEL_SRC}/MockChannel.cpp"

Testing/Stubs/Inc/tim.h
Penguronik: Yup was just leftover from before we decided to make the custom unit test modelconfig

Testing/Stubs/Inc/main.h
Penguronik: Yep but just would need some stuff like TIM_HandleTypeDef as long as the PWMChannel header is still compiling, mentioned issue in comment about mocking MockChannel as well

Testing/Stubs/Inc/FreeRTOS.h
Penguronik: AM is using TickType_t, portMAX_DELAY, and pdPASS in semaphore related stuff right now

AttitudeManager/Src/AM.cpp
Penguronik: I would still need to friend the class though as that is how I find out what the motors are being set to, I could fix this problem and [this problem](https://github.com/UWARG/ZeroPilot-SW-3.5/pull/31/files#r1367636496) at the same time by creating a global array that holds motor outputs (kinda simulating how the stm sets regkeys for motor outputs is my thinking)

Penguronik: Actually after trying the global array way, it seems like I would still need to friend the class as outputToMotor is a private method that AM is supposed to call on its own so I don't know how else I could test it

AttitudeManager/Tests/AMTest.cpp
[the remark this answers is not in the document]
Aidan-B: > My only problem with this is that as the driverConstructor returns a pointer to MotorChannel base class, I'm not able to access any of the variables or methods I created inside my MockMotor class for testing purposes (such as a way to display the value it would normally set in __HAL_TIM_SET_COMPARE)

Don't use the configs for testing your class here. Create an instance of a mock motor channel and pass it to the constructor of AM, as mentioned (dependency injection).

AttitudeManager/Src/AM.cpp
Aidan-B: > I would still need to friend the class though as that is how I find out what the motors are being set to, I could fix this problem and [this problem](https://github.com/UWARG/ZeroPilot-SW-3.5/pull/31/files#r1367636496) at the same time by creating a global array that holds motor outputs (kinda simulating how the stm sets registers for motor outputs is my thinking)

You should not need to friend AM to inspect the value that a motor was set to, as the test will have the motor channel in its local scope to inspect.

AttitudeManager/Tests/AMTest.cpp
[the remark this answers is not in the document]
Aidan-B: Please write a comment to succinctly explain this

AttitudeManager/Tests/AMTest.cpp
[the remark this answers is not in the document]
Aidan-B: > Also placing the mock alongside the other motor types makes it that I have to compile PWMChannel too even though I'm not using it, which gives me compiler errors for missing SystemCoreClock, __HAL_TIM_SET_PRESCALER, etc which I can just stub but feels like I shouldn't need to. Not sure if it would be an acceptable solution to just change cmakelists from "${MOTORCHANNEL_SRC}/*.cpp" to "${MOTORCHANNEL_SRC}/MockChannel.cpp"

I would suggest trying then to place all of the mocks in their own folder, and use the original includes then? You're right that we shouldn't need to compile these other files that aren't being used.

AttitudeManager/Inc/AM.hpp
```diff
@@ -20,17 +21,13 @@
...
-// End of test class forward declarations
-
 class AttitudeManager {
    public:
     static void setControlInputs(const AttitudeManagerInput& new_control_inputs);
 
     static AttitudeManagerInput getControlInputs();
 
-    AttitudeManager(Flightmode* control_algorithm);
+    AttitudeManager(Flightmode* control_algorithm, config::Motor_t motors[]);
```
Aidan-B, line 30: Using `MotorChannel` would be preferable here

Drivers/MotorChannel/Inc/MockChannel.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef ZP_D_MOCK_CHANNEL_HPP_
+#define ZP_D_MOCK_CHANNEL_HPP_
+
+#include "main.h"
+
+#include "ZP_D_MotorChannel.hpp"
+
+namespace mock {
+extern uint8_t *motorOutputs;
+}
```
Aidan-B, line 10: Just make this a member of the `MockChannel` class

Tools/Testing/CMakeLists.txt
```diff
@@ -83,14 +83,17 @@ file(GLOB TM_CODE
...
 
-include_directories(${DRIVERS_INC})
+include_directories(${DRIVERS_INC} ${MOTORCHANNEL_INC})
 file(GLOB DRIVERS_CODE
     "${DRIVERS_FOLDER}/Tests/*.cpp"
 
     # add necessary Drivers source files below
     # "${DRIVERS_SRC}/foo.cpp"
     "${DRIVERS_FOLDER}/Common/Src/circular_buffer.cpp"
+    "${MOTORCHANNEL_SRC}/*.cpp"
```
Aidan-B, line 96: You can fix the build issues you mentioned by either specifying the mock you wish to build, or creating a mock folder

Aidan-B (changes requested):

antimonyB (commented):
Looks like this will be good once you reincorporate the commented-out tests and address Aidan's suggestions.

Aidan-B:
Take a look at https://google.github.io/googletest/reference/mocking.html. It allows you to very easily create a mock, and then use a bunch of methods to check that the methods have been called multiple times.

Penguronik:
Just pushed a commit that should resolve all conversations, will look over everything again and resolve conversations tomorrow, also will copy over the now SM portion to a new branch before we squash and merge. Should still be good to review though, just not ready to merge just yet.

Testing/Stubs/Inc/main.h
Penguronik: Hey so I moved the mocks to a mock folder but since I am still compiling the headers from MotorChannel folder for the MotorChannel.hpp that also causes me to compile PWMChannel.hpp which requires the stuff in main.h. Is there any smarter way to do this, since PWMChannel isn't actually required in unit tests but its in the same folder as MotorChannel

AttitudeManager/Tests/AMTest.cpp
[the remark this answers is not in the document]
Penguronik: Hey so I still have this test in there, should I? or I guess is there a better way to indicate I am testing for the function to not segfault. I can't really use EXPECT_CALL as I did in other tests because there is no MockChannel I can look at when there are no motors

Testing/Stubs/Inc/main.h
Penguronik: Turns out hpp files are only compiled if they are #included from another file. temp_drivers was including PWM but it seems like its enough to just include motors so that's what I switched it to

Aidan-B (approved):
I'll trust that my previous review was fine... I remember it being very close just needing some massaging with the tests, which appear to have happened. Sorry I took forever to get to this review

AttitudeManager/Src/AM.cpp
```diff
@@ -35,13 +35,35 @@ AttitudeManagerInput AttitudeManager::getControlInputs() {
...
-    control_algorithm->run();
+    controlAlgorithm_->run();
 
     // Write motor outputs
 }
 
+void AttitudeManager::outputToMotor(config::ControlAxis_t axis, uint8_t percent) {
+    // Move through the portion of the motorInstances array that matches the wanted axis.
+    // The motorReferences array holds references to the wanted positions in the motorInstances array
+    for (uint8_t motorCount{0}; motorCount < numMotorsPerAxis_[axis]; motorCount++) {
```
antholuo, line 59:
```suggestion
    for (uint8_t motorCount{0}; motorCount < numMotorsPerAxis_[axis]; ++motorCount) {
```

antholuo (approved):

AttitudeManager/Src/AM.cpp
antholuo: For more info/context: https://stackoverflow.com/questions/24853/what-is-the-difference-between-i-and-i#:~:text=%2B%2Bi%20increments%20the%20value,i%20%2C%20as%20it's%20slightly%20faster.

PR #43, opened 2024-03-30
Moving Average Filter Implementation

twilkhoo:
## Description

Adding a moving average filter, which should be used to preprocess the data going into our AHRS solution.

## Testing

Todo

Working on gTesting right now

## Documentation

*Milestone number and name:*
Simple IMU Data filters, M3

*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1206513612014108/f

Todo

## Reminders

- [ ] Add reviewers to the PR


AttitudeManager/Src/MovingAverage.cpp
```diff
@@ -0,0 +1,41 @@
...
+Window<T>::Window() {
+  memset(window, 0, MAX_WINDOW_SIZE);
+  size = 0;
+  pos = 0;
+}
+
+template<typename T>
+void Window<T>::update(T newValue) {
+  window[pos] = newValue;
+  pos++;
```
antholuo, line 14: you can post-increment within your array access I believe:

https://stackoverflow.com/questions/34250975/increment-operator-inside-array

AttitudeManager/Src/MovingAverage.cpp
```diff
@@ -0,0 +1,41 @@
...
+    return FAILURE;
+  }
+
+  int sum = 0;
+  for (int i = pos; i > pos - n; i--) {
+    uint8_t realIdx = pos % MAX_WINDOW_SIZE;
+    sum += window[realIdx];
+  }
+
+  *result = static_cast<T>(sum) / n;
```
antholuo, line 39: I think the idea for the static cast is good - the data that IMU's deal with is almost always floating point decimal (i.e. 0.9g, 1.1g, 1.7137g, etc), so it might be worth forcing a float return.

Something about safety-critical applications and never "guessing" your return type. Search for "cast" in this doc: https://embeddedgurus.com/barr-code/category/coding-standards/page/2/ as a starting point

antholuo (changes requested):
Overall looks good - just some small changes.

I'd suggest moving all the filtering metatypes into "common" folder, and then keep the IMU stuff inside AM for now.

Would **love** if you could test this in person to see it working! Then get some code profiling going, I see lots and lots of math & memory operations, and wonder if we could improve our performance by using DMA or TCM better :D . That can be a second PR though, don't try to push it all in one.

Overall super great though! very excited to see this

PR #47, opened 2024-04-15
Added altimeter .hpp and .cpp files

FolaFatola:
## Description

Added altimeter driver .hpp and .cpp files.

Reads temperature and pressure and calculates height in meters.

## Testing

Printed values to terminal using UART.

## Documentation
*Milestone number and name:*
M4

*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204645027312243

### DIFF
--- a/Altimeter/Inc/altimeter.hpp
+++ b/Altimeter/Inc/altimeter.hpp
@@ -19,7 +19,6 @@ constexpr uint8_t PROM_READ_ADDRESS_5 = 0xAA;
 constexpr uint8_t PROM_READ_ADDRESS_6 = 0xAC;
 constexpr uint8_t PROM_READ_ADDRESS_7 = 0xAE;
 
-
 // digital pressure/temperature commands at different sampling rates.
 constexpr uint8_t CONVERT_D1_OSR_256 = 0x40;
 constexpr uint8_t CONVERT_D1_OSR_512 = 0x42;
@@ -34,16 +33,14 @@ constexpr uint8_t CONVERT_D2_OSR_4096 = 0x58;
 
 constexpr uint8_t ADC_READ = 0x0;
 
-constexpr uint32_t TIMEOUT = 1 << 31;
 
 class MS5611{
 public:
 	/**
-	* Constructor for the AltDevice class.
+	* Constructor for the MS5611 class.
 	* It assigns the spi_handler and ports and pins related to the chip select and protocol pins.
-	* This function sends the reset command, populates the cal_coeffs
-	* struct members with the calibration coefficients in the sensor PROM,
-	* and gets the current elevation above sea level.
+	* This function resets the device, populates reads calibration data from the device's PROM
+	* PROM, and gets the sensor's initial elevation above sea level.
 	 *
 	*
 	* @param spi_handle -> hspix handle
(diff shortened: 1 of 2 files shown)


Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	coeffs_.t_ref = promRead(spi_handle, cs_port, cs_pin, PROM_READ_ADDRESS_5);
+	coeffs_.temp_sens = promRead(spi_handle, cs_port, cs_pin, PROM_READ_ADDRESS_6);
+}
+
+
+uint16_t AltDevice::promRead(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, uint16_t cs_pin, uint8_t address){
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_RESET);
+
+	uint8_t send_read_command[3] = {address, 0xFF, 0xFF};
```
antholuo, line 51: if promRead is used often, it would be faster to put this variable as a static const (most compilers will optimize it a bit then). Depends on where your program ends up placing this function in memory.

Altimeter/Inc/altimeter.hpp
```diff
@@ -0,0 +1,166 @@
...
+public:
+	/**
+	* Constructor for the AltDevice class
+	 *
+	* @param ps_port -> protocol select port
+	* @param ps_pin -> protocol select pin
+	 *
+	* @return none
+	 */
+	AltDevice(GPIO_TypeDef *ps_port, uint16_t ps_pin);
```
HardyYu, line 64: In my opinion, a driver code could be a lot easier to use if we only initiate the device once with the SPI, cs, ps line bundled with this driver code so that the user of your driver doesn't have to know the SPI it is using every time they call get_altitude function. That means I think `SPI_HandleTypeDef` and `GPIO_TypeDef` they should be the private variable of this class.

Altimeter/Src/altimeter.cpp
antholuo: also - just in the vein of "making things clear", "send_read_command" doesn't make much sense(?) "read_command" or "spi_tx_data" might be better options

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	uint8_t coefficient_val[3];
+
+	HAL_SPI_TransmitReceive(spi_handle, send_read_command, coefficient_val, 3, TIMEOUT);
```
antholuo, line 54: I say "send_read_command" might not be great because I was reading this function and had no clue what "send_read_command" was doing in this section. 

I would prefer seeing `HAL_SPI_TransmitReceive(spi_handle, tx_data, rx_data, data_size, TIMEOUT)`

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_SET);
+
+	//Grab conversion coefficients.
```
antholuo, line 58:
```suggestion
	// Grab conversion coefficients.
```

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+uint32_t AltDevice::uncompensatedPressureTemperature(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, uint16_t cs_pin, uint8_t conversion_command){
+
+	uint8_t digital_pres_temp[1] = {conversion_command};
+	uint8_t receive_buffer[1];
+
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_RESET);
+
+	/* send the digital pressure or temperature with the command */
+	HAL_SPI_TransmitReceive(spi_handle, digital_pres_temp, receive_buffer, 1, TIMEOUT);
```
HardyYu, line 75: I think this function could be changed to an interrupt based version of SPI receive in the next revision, so when the worst situation happened, we are not waiting for this timeout forever. Every return value from the communication function call to an external device should be checked. The failed communication should be handled explicitly, and inform to the function caller.

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	uint16_t conversion_coefficient = (coefficient_val[1] << 8) | (coefficient_val[2]);
+
+	return conversion_coefficient;
+}
+
+
+
+uint32_t AltDevice::uncompensatedPressureTemperature(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, uint16_t cs_pin, uint8_t conversion_command){
+
+	uint8_t digital_pres_temp[1] = {conversion_command};
```
antholuo, line 68: why do we need curly braces {}

also...why does this need to be a uint8_t array of 1 depth

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_RESET);
+
+	/* send the digital pressure or temperature with the command */
+	HAL_SPI_TransmitReceive(spi_handle, digital_pres_temp, receive_buffer, 1, TIMEOUT);
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_SET);
+
+	/* at least 8.22 ms delay required for ADC conversion.*/
+	osDelay(9);
+
+	uint8_t tx_buffer[4] = {ADC_READ, 0xFF, 0xFF, 0xFF};
```
HardyYu, line 81: this variable is never changed, please add `const` and capitalize this variable

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	return conversion_coefficient;
+}
+
+
+
+uint32_t AltDevice::uncompensatedPressureTemperature(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, uint16_t cs_pin, uint8_t conversion_command){
+
+	uint8_t digital_pres_temp[1] = {conversion_command};
+	uint8_t receive_buffer[1];
```
antholuo, line 69: same here , why are we making an array of depth 1?

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_RESET);
+
+	/*Start adc conversion to get value */
+	HAL_SPI_TransmitReceive(spi_handle, tx_buffer, rx_buffer, 4, TIMEOUT);
+
+	HAL_GPIO_WritePin(cs_port, cs_pin, GPIO_PIN_SET);
+
+	//get 24-bit uncompensated pressure or temperature value value.
+	uint32_t uncompensated_value = (rx_buffer[1] << 16 | rx_buffer[2] << 8 | rx_buffer[3]);
```
HardyYu, line 92: also do a `& 0x00FFFFFF` at the end?

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+			sens2 = sens2 + 11 * pow((temp + 1500), 2) / 2;
+		}
+
+		temp = temp - t2;
+		off = off - off2;
+		sens = sens - sens2;
+
+	}
+
+	float pressure = (digital_pressure * (sens / 2097152.0f) - off) / 32768.0f;
```
HardyYu, line 129: A lot of magic numbers here in this function, it will be the best if each of them can be assigned with a name. No need for the small whole numbers such as `2`, `5`.

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	return temp_;
+}
+
+float AltDevice::getAltAboveSeaLvl(){
+
+	float reference_temp = 288.15f; /* Ref temperature in Kelvins */
+	float temp_lapse_rate = 0.0065f;
+	float exp_gmrl = 5.2558;
+
+	float reference_pressure = 101325.0f; /* in Pa */
```
HardyYu, line 152: yeah like what you are doing here is awesome

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	uint32_t digital_temperature = uncompensatedPressureTemperature(spi_handle, cs_port, cs_pin, CONVERT_D2_OSR_256);
+	uint32_t digital_pressure = uncompensatedPressureTemperature(spi_handle, cs_port, cs_pin, CONVERT_D1_OSR_256);
+
+	float dt = digital_temperature - (float)coeffs_.t_ref * 256.0f;
+	float temp = 2000 + dt * (float) coeffs_.temp_sens / pow(2, 23);
+
+
+	float off = coeffs_.off * 65536.0f + (coeffs_.tco * dt) / 128.0f;
+	float sens = coeffs_.sens * 32768.0f  + (coeffs_.tcs * dt) / 256.0f;
```
antholuo, line 109: Descriptive naming here too please!

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	/*Second order temperature compensation */
+	if(temp < 2000){
```
antholuo, line 112:
```suggestion
	if (temp < 2000) {
```

Take a look at warg's clang-format if you haven't already, it should take care of stuff like this

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+		sens = sens - sens2;
+
+	}
+
+	float pressure = (digital_pressure * (sens / 2097152.0f) - off) / 32768.0f;
+	pressure /= 100.0f;
+	temp /= 100.0f;
+
+	pres_ = pressure;
+	temp_ = temp;
```
antholuo, line 134: Comments for magic numbers would be helpful.

In general, we try and avoid magic numbers and instead use consts with clear names!

HardyYu (commented):
Good stuff! Can't wait to see it get used in ZP

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+	float exponent = (log(current_pressure) - log(reference_pressure)) / exp_gmrl;
+
+	float height = reference_temp/temp_lapse_rate * (1 - pow(M_E, exponent));
+
+	return height;
+}
+
+
+float AltDevice::getAltitude(){
+	height_ = getAltAboveSeaLvl();
```
antholuo, line 163: is height_ a private var? Why are we modifying a private var here but not in the above function?

antholuo (changes requested):
Overall looks pretty good and solid, very pleased with what we have so far.

Some formatting changes requested (take a look at WARG's clang-format script).

Also wanted to make sure we have clear variable names, and are managing our classes / structs / data properly in order to make future-us lives easier!

edit: Also try and include full test-methodology and screenshots/photos in the PR! You want to make it as easy as possible for someone else to re-create your test and verify results, be it now or 5 months in the future

Altimeter/Src/altimeter.cpp
FolaFatola: Okay

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void AltDevice::altInit(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, uint16_t cs_pin){
+	reset(spi_handle, cs_port, cs_pin);
+	getConvCoeffs(spi_handle, cs_port, cs_pin);
+
+	/*Get the average of 100 readings*/
+
+	float reading_sum = 0.0f;
+	for(int i = 0; i < 100; i++){
+		calculateTempPres(spi_handle, cs_port, cs_pin);
```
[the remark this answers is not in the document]
FolaFatola: No. I've made the adjustments. You'll see them in the upcoming pr.

Altimeter/Src/altimeter.cpp
FolaFatola: Done

PR #48, opened 2024-04-18
IMU class template

antholuo:
Adds a IMU class template.

Largely lifted from ZP 1.0.

Need to know the following:
- How we want to return / express error states
- Whether or not we want to keep singleton architecture
- Where hardware SPI/I2C handles are defined
- How to create initializer and whether or not we want public `init` functions.
- How many folders we want under /drivers/

### DIFF
--- a/Common/Inc/CommonDataTypes.hpp
+++ b/Common/Inc/CommonDataTypes.hpp
@@ -1,21 +1,35 @@
-//
-// Created by Gordon Fountain 2022-11-30.
-//
+/** CommonDataTypes.hpp
+ * Contains a bunch of common datatypes!
+ * 
+ * Created by: Gordon Fountain 2022-11-30
+ * Updated by: Anthony Luo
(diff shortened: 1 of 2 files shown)


Common/Inc/CommonDataTypes.hpp
```diff
@@ -1,21 +1,35 @@
...
 #include <stdint.h>
 
+/************************************************
+ * Sensor/Error structs (can be either or. tbd ig)
+ ************************************************/
+typedef enum SensorErrorCode {
+    SENSOR_SUCCESS = 0,
+    SENSOR_FAIL = -1,
+    SENSOR_BUSY = 1,
+} SensorErrorCode;
```
HardyYu, line 21: This enum name doesn't follow the style guide: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards

HardyYu (changes requested):

PR #49, opened 2024-04-29
MS5611 Driver Update

FolaFatola:
## Description

Comments made in a previous PR were addressed.
PR: https://github.com/UWARG/efs-zeropilot-3.5/pull/47

## Testing

Temperature, pressure and altitude values were printed to the UART. For the altitude, the barometer was raised a few meters and measurements were observed. For temperature, values were compared to local thermometer. The pressure was compared to the pressure of the city I was in (googled).

<br/>
height
altitude

**Testing for failures:**

The measurements in this driver return a bool - true if no communication errors occur and false if they do occur
image

If HAL_SPI_TransmitReceive_IT doesn't HAL_OK, communication_success_ is assigned the value false. To test if that would cause if that would cause the get functions to return false (getPressure, getTemperature, etc), I would assign communication_success_ false and check the results.

Forced failure example #1:
image

Forced failure example #2:
image

Results:
image

Test loop:
image

<br/>

<br/>

## Documentation

*Milestone number and name:*

*Link to Asana task:*

### DIFF
--- a/Altimeter/Src/altimeter.cpp
+++ b/Altimeter/Src/altimeter.cpp
@@ -8,19 +8,13 @@ MS5611::MS5611(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, GPIO_TypeDe
 	cs_pin_ = cs_pin;
 	ps_pin_ = ps_pin;
 	HAL_GPIO_WritePin(ps_port_, ps_pin_, GPIO_PIN_RESET);
-}
-
-void MS5611::altInit(){
-
 
 	reset();
 	getConvCoeffs();
 
-	/*Get the average of 100 readings*/
-
+	// Get the average of 100 readings.
 	float reading_sum = 0.0f;
 	for(int i = 0; i < 100; i++){
-		calculateTemperatureAndPressure();
 		reading_sum += getAltitudeAboveSeaLevel();
 	}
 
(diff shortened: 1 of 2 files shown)


Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,253 @@
...
+void MS5611::reset(){
+	HAL_GPIO_WritePin(cs_port_, cs_pin_, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(cs_port_, cs_pin_, GPIO_PIN_RESET);
+
+	uint8_t tx_data[3] = {RESET_COMMAND, 0xFF, 0xFF};
+	uint8_t rx_data[3];
+	uint16_t data_size = 3;
+
+	spi_status_ = HAL_SPI_TransmitReceive_IT(spi_handle_, tx_data, rx_data, data_size);
+	if (spi_status_ != HAL_OK) {
```
HardyYu, line 49: It seems to me the `spi_status_` in this class could totally be initialized locally in every function instead of making it a private member variable.

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,253 @@
...
+	// Subtract this from measurements at greater heights for the height above that initial surface.
+	float base_elevation = 0.0f;
+	if (communication_success_) {   //if HAL_SPI_TransmitReceive_IT doesn't fail in reset or getPromData()
+
+		float altitude_above_sea_level_measurement = 0;
+		int num_successful_measurements = 0;
+		for (int i = 0; i < 100; i++) { 			//take a few measurements to make sure they are accurate/consistent.
+			communication_success_ = true;
+			communication_success_ = getAltitudeAboveSeaLevel(altitude_above_sea_level_measurement);
+			osDelay(10);
```
HardyYu, line 27: why the delay is 10 here, but 9 elsewhere? Especially this operation will be repeated 100 times in the initialization.

HardyYu (approved):
Great work Fola! I have some minor comments but I don't think those problems are big enough to stop you from merging.

antholuo (approved):
Sorry for the late review - but it all looks good to me.

Thanks so much for the good work, testing looks great.

PR #53, opened 2024-05-22
Drivers/altimeter

FolaFatola:
## Description
Update of MS5611 Driver. Function return types changed from bool to float with FLT_MAX used to signify errors. Communication_success_ internal variable removed and replaced with other internal variables.
<br/>

Temperature, pressure and altitude values were printed to the UART. For the altitude, the barometer was raised a few meters and measurements were observed. For temperature, values were compared to local thermometer. The pressure was compared to the pressure of the city I was in (googled).

## Testing
image

Screen Shot 2024-05-21 at 11 44 12 PM

<br/>

<br/>

## Reminders

- [ ] Add reviewers to the PR

### DIFF
--- a/Altimeter/Src/altimeter.cpp
+++ b/Altimeter/Src/altimeter.cpp
@@ -8,19 +8,13 @@ MS5611::MS5611(SPI_HandleTypeDef *spi_handle, GPIO_TypeDef *cs_port, GPIO_TypeDe
 	cs_pin_ = cs_pin;
 	ps_pin_ = ps_pin;
 	HAL_GPIO_WritePin(ps_port_, ps_pin_, GPIO_PIN_RESET);
-}
-
-void MS5611::altInit(){
-
 
 	reset();
 	getConvCoeffs();
 
-	/*Get the average of 100 readings*/
-
+	// Get the average of 100 readings.
 	float reading_sum = 0.0f;
 	for(int i = 0; i < 100; i++){
-		calculateTemperatureAndPressure();
 		reading_sum += getAltitudeAboveSeaLevel();
 	}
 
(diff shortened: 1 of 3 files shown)


Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+bool MS5611::getBarometer(uint8_t conversion_command, uint32_t &uncompensated_pressure_temperature){
+
+	if(!setup_done_){
+		return false;
+	}
+
+	uint8_t tx_data = conversion_command;
+	uint8_t rx_data;
+	uint8_t tx_data_2[4] = {ADC_READ, 0xFF, 0xFF, 0xFF};
```
HardyYu, line 81: Looks like this can be a const. Same with the other tx_data.

Altimeter/Src/altimeter.cpp
```diff
@@ -0,0 +1,229 @@
...
+	uint8_t rx_data_2[4];
+	uint16_t data_size = 1;
+
+	HAL_GPIO_WritePin(cs_port_, cs_pin_, GPIO_PIN_SET);
+	HAL_GPIO_WritePin(cs_port_, cs_pin_, GPIO_PIN_RESET);
+
+	// send the digital pressure or temperature with the command.
+	HAL_StatusTypeDef spi_status_ = HAL_SPI_TransmitReceive_IT(spi_handle_, &tx_data, &rx_data, data_size);
```
HardyYu, line 89: Sorry, it was me in the previous PR saying maybe we could use interrupt-based data transceiving in the future, but I actually meant using DMA. And here I think it makes sense to use normal polling since you are reading the transmitting at the same time. I thought if you call the hal function ended with `_IT`, you are also triggering an ISR callback function, but you also don't need to use it here, so there is not a big difference between using interrupt base or pulling based transceiving.

HardyYu (approved):
Well done Fola! It looks so good. I still left some comments here but they are all minor, and they will not change the fact that this code will work.

PR #55, opened 2024-05-27
Feature/circular queue

RRahul2004:
## Description

Implemented TM Cicular Queue

<br/>

to allow Tm to store MAVLINK message in queue waiting for use

<br/>

## Testing

<br/>

<br/>

## Documentation

*Milestone number and name:*

*Link to Asana task:*

<br/>

## Reminders

- [ ] Add reviewers to the PR

### DIFF
--- a/Drivers/common/circular_buffer/inc/circular_buffer.hpp
+++ b/Drivers/common/circular_buffer/inc/circular_buffer.hpp
@@ -62,6 +62,13 @@ class CircularBuffer {
      */
     uint16_t getFreeSpaceBytes();
 
+    /**
+     * @brief Provide the size of the buffer.
+     *
+     * @return uint16_t The size of the buffer.
+     */
+    uint16_t getSize_();
+
    private:
     uint8_t* buf_;
     uint16_t size_;
--- a/Drivers/common/circular_buffer/inc/circular_buffer.hpp
+++ b/Drivers/common/circular_buffer/inc/circular_buffer.hpp
@@ -16,6 +16,16 @@ class CircularBuffer {
      */
     CircularBuffer(uint8_t* buf, uint16_t size);
 
+    /** @brief Copy Constructor for the CircularBuffer class
+     *
+     *  CircularBuffer will create a copy of  a stack-allocated byte array as a
+     *  circular buffer.
+     *
+     *  .
+     *  @param other an instance of a circular buffer 
+     */
+    CircularBuffer(const CircularBuffer& other);
+
     /** @brief Read the value of a byte in the circular buffer
      *          without incrementing the read pointer
      *
@@ -69,7 +79,7 @@ class CircularBuffer {
      */
     uint16_t getSize();
 
-   private:
+   protected:
     uint8_t* buf_;
     uint16_t size_;
     uint16_t writePtr_ = 0;
--- a/Drivers/common/circular_buffer/inc/circular_buffer.hpp
+++ b/Drivers/common/circular_buffer/inc/circular_buffer.hpp
@@ -22,7 +22,7 @@ class CircularBuffer {
      *  circular buffer.
      *
      *  .
-     *  @param other an instance of a circular buffer 
+     *  @param other an instance of a circula   r buffer 
      */
     CircularBuffer(const CircularBuffer& other);
 
@@ -72,12 +72,6 @@ class CircularBuffer {
      */
     uint16_t getFreeSpaceBytes();
 
-    /**
-     * @brief Provide the size of the buffer.
-     *
-     * @return uint16_t The size of the buffer.
-     */
-    uint16_t getSize();
 
    protected:
     uint8_t* buf_;
--- a/Drivers/common/circular_buffer/inc/circular_buffer.hpp
+++ b/Drivers/common/circular_buffer/inc/circular_buffer.hpp
@@ -67,7 +67,7 @@ class CircularBuffer {
      *
      * @return uint16_t The size of the buffer.
      */
-    uint16_t getSize_();
+    uint16_t getSize();
 
    private:
     uint8_t* buf_;
--- a/Drivers/common/circular_buffer/src/circular_buffer.cpp
+++ b/Drivers/common/circular_buffer/src/circular_buffer.cpp
@@ -7,6 +7,10 @@ CircularBuffer::CircularBuffer(uint8_t* buf, uint16_t size) {
     size_ = size;
 }
 
+CircularBuffer::CircularBuffer(const CircularBuffer& other) {
+  
+}
+
 bool CircularBuffer::peek(uint8_t& byte, uint16_t index) {
     bool success = true;
 
@@ -70,7 +74,3 @@ bool CircularBuffer::write(uint8_t byte) {
 uint16_t CircularBuffer::getFreeSpaceBytes() { 
     //NOT IMPLEMENTED YET
     return -1; }
-
-uint16_t CircularBuffer::getSize(){
-    return  size_;
-}
--- a/Drivers/common/circular_buffer/src/circular_buffer.cpp
+++ b/Drivers/common/circular_buffer/src/circular_buffer.cpp
@@ -71,6 +71,6 @@ uint16_t CircularBuffer::getFreeSpaceBytes() {
     //NOT IMPLEMENTED YET
     return -1; }
 
-uint16_t CircularBuffer::getSize_(){
+uint16_t CircularBuffer::getSize(){
     return  size_;
 }
--- a/Drivers/common/circular_buffer/src/circular_buffer.cpp
+++ b/Drivers/common/circular_buffer/src/circular_buffer.cpp
@@ -70,3 +70,7 @@ bool CircularBuffer::write(uint8_t byte) {
 uint16_t CircularBuffer::getFreeSpaceBytes() { 
     //NOT IMPLEMENTED YET
     return -1; }
+
+uint16_t CircularBuffer::getSize_(){
+    return  size_;
+}
--- a/TelemetryManager/Inc/TMCircularBuffer.hpp
+++ b/TelemetryManager/Inc/TMCircularBuffer.hpp
@@ -78,7 +78,7 @@ class TMCircularBuffer : public CircularBuffer {
     int currentIndex();
 
     private: 
-    int index = 0;
+    uint8_t index_ = 0;
 };
 
 #endif
--- a/TelemetryManager/Inc/TMCircularBuffer.hpp
+++ b/TelemetryManager/Inc/TMCircularBuffer.hpp
@@ -14,6 +14,7 @@
 
 #ifndef TMCIRCULARBUFFER_H
 #define TMCIRCULARBUFFER_H
+#define MAX_MAVLINK_MSG_SIZE 280
 
 #include "circular_buffer.hpp"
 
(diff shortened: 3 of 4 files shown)


TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
 
-#include "CircularBuffer.hpp"
+#include "TMCircularBuffer.hpp"
 
-TMCircularBuffer::TMCircularBuffer(uint8_t* buf, uint16_t size): CircularBuffer(buf, size) {
+TMCircularBuffer::TMCircularBuffer(uint8_t* buf, uint16_t size): CircularBuffer(*buf, size) {
     // Constructor
```
HardyYu, line 5: Things about cpp inheritance, and the way how you inherit from CircularBuffer will not make your TMCircularBuffer work. 
If your class tries to inherit the private variables from the superclass, you won't get access to it. And I think a lot of methods depend on those private variables. Should change the parent class private variables to protected to allow your class to keep using them.
https://www.geeksforgeeks.org/inheritance-in-c/

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
 }
 
-TMCircularBuffer::TMCircularBuffer(CircularBuffer buf): CircularBuffer(buf) {
+TMCircularBuffer::TMCircularBuffer(CircularBuffer* buf): CircularBuffer(*buf) {
```
HardyYu, line 8: I don't see there is a constructor in the parent class that does `CircularBuffer::CircularBuffer(CircularBuffer* buf) `

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
     // Constructor
 }
 
 TMCircularBuffer::~TMCircularBuffer() {
     // Destructor
-}
+    delete[] buf;
```
HardyYu, line 14: The buf you are trying to delete here doesn't exist.

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
+    buf = nullptr;
```
HardyYu, line 15: The deconstruction to the `buf_` shouldn't be handled by this class since the original buffer was passed by a pointer. If you deconstruct it here, the outer buffer pointer will also be deconstructed. There is also a good chance that the outer buffer isn't even dynamically allocated, in that case, deconstruct them here doesn't make any sense.

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
+            return index;
+        }
+        index = (index + 1) % size_;
+        count ++;
+    }
     return -1;
 }
 
-int TMCircularBuffer::currentIndex() { return -1; }
+int TMCircularBuffer::currentIndex() { return  readPtr_; }
```
HardyYu, line 56: Yeah... this won't work, because this class doesn't have the access to the private variables in its parent class

HardyYu (changes requested):
Good try! I listed some points that could cause this class to not work, please review and change them. Don't be afraid of changing someone else's code! Sometimes other people's code is faulted or no long apply to the current use case anymore. I also recommend to include your header in the `Boardfiles/nucleol552zeq/Core/Src/main.cpp` to check whether your code can pass compilation. The way how compiler works is that if you don't use it, then it won't be compiled. You can do some testing in the main to check whether or not your code is actually working.

TelemetryManager/Src/TMCircularBuffer.cpp
Yaremadzulynsky: This was my fault. I asked Rahul to make a last-minute change.

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
 }
 
 TMCircularBuffer::~TMCircularBuffer() {
     // Destructor
-}
+    delete[] buf;
+    buf = nullptr;
+ }
+
+TMCircularBuffer::bool TMCircularBuffer::dequeue(MAVLinkByte &byte) {
```
Yaremadzulynsky, line 18: One thing you could do to manipulate/retrieve info from the buffer without needing to access the private variable, is to use the already-made functions in the base class. For example, your enqueue() function contents could be something like 
{
read(youre byte, 1);
}

Since the base class already has most circular buffer methods implemented, to save yourself work, you can just use them.

Also, dequeue should return the dequeued byte so we can use it, not a boolean.

Drivers/common/circular_buffer/src/circular_buffer.cpp
```diff
@@ -69,4 +69,8 @@ bool CircularBuffer::write(uint8_t byte) {
 
 uint16_t CircularBuffer::getFreeSpaceBytes() { 
     //NOT IMPLEMENTED YET
-    return -1; }
\ No newline at end of file
+    return -1; }
```
HardyYu, line 72: shouldn't return -1 for "uint16_t" type ...

Drivers/common/circular_buffer/src/circular_buffer.cpp
```diff
@@ -69,4 +69,8 @@ bool CircularBuffer::write(uint8_t byte) {
...
+
+uint16_t CircularBuffer::getSize_(){
```
HardyYu, line 74: could you rename this method to be `getSize` to get the consistency with other methods?

TelemetryManager/Inc/TMCircularBuffer.hpp
```diff
@@ -76,6 +76,9 @@ class TMCircularBuffer : public CircularBuffer {
      * @return int The index of the current byte in the queue.
      */
     int currentIndex();
+
+    private: 
+    int index = 0;
```
HardyYu, line 81: name it to be `index_`? If the index could never to be less than 0, then probably consider change the type to some unsigned data type.

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
+   while(count < size){
+      if( buf[index] == 0xFD){
+        return index;
+      }
+      
+      index = (index + 1) % size;
+      count ++;
+
+   }
+
```
HardyYu, line 51: this method is supposed to return an `int` type data

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
     // Destructor
+   
+ }
+
+TMCircularBuffer::MAVLinkByte  TMCircularBuffer::dequeue() {
+    uint8_t* res;
+    if(read(res,1)){
+        return *res;
+    }
+    return *res;
```
HardyYu, line 22: return data type inconsistent with a function declaration. `uint8_t` and `MAVLinkByte`

HardyYu (changes requested):
Nice! Yea definitely better than last time. Left some comments to be addressed.

TelemetryManager/Src/TMCircularBuffer.cpp
HardyYu: this function is still questionable? Can you confirm that this will work?

TelemetryManager/Src/TMCircularBuffer.cpp
HardyYu: Shouldn't have line 22 return *res since you failed to read from circular buffer, probably return `0` instead.

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
 }
 
-TMCircularBuffer::MAVLinkByte TMCircularBuffer::dequeue() { return 0; }
-
-void TMCircularBuffer::enqueue(MAVLinkByte byte) {
+bool TMCircularBuffer::enqueue(MAVLinkByte byte) {
     // Enqueue the byte
+    if(write(byte)){
+        index ++;
```
HardyYu, line 28: index = (index + 1) % getSize_();  // Wrap around index to prevent overflow

HardyYu (changes requested):

TelemetryManager/Src/TMCircularBuffer.cpp
RRahul2004: function returns -1 if not found

TelemetryManager/Inc/TMCircularBuffer.hpp
```diff
@@ -76,6 +76,9 @@ class TMCircularBuffer : public CircularBuffer {
      * @return int The index of the current byte in the queue.
      */
     int currentIndex();
+
+    private: 
+    uint8_t index_ = 0;
```
Yaremadzulynsky, line 81: What happens if the read method is used from the base class? The index_ in TMCircularBuffer would not track that right? The RFD900 DMA uses the base class methods, but we access the same buffer via creating a TMCircularBuffer with the common CircularBuffer object. Maybe you could have a getIndex() method in the common CircularBuffer and just have the TMCircularBuffer currentIndex function return that?

TelemetryManager/Src/TMCircularBuffer.cpp
Yaremadzulynsky: @RRahul2004 are you able to do some manual testing to confirm this works?

TelemetryManager/Src/TMCircularBuffer.cpp
```diff
@@ -1,30 +1,56 @@
...
     // Enqueue the byte
+    if(write(byte)){
+        index_ = (index_ + 1) % getSize();
+        return true;
+    }
+     return false;
 }
 
-int TMCircularBuffer::lastFullMessageEndIndex() {
+int TMCircularBuffer::bytesUntilLastMessageEnd(uint8_t* buf) {
```
Yaremadzulynsky, line 34: If you make buf in the common circular buffer protected, you should be able to access it in this method without requiring it as a parameter.

Drivers/common/circular_buffer/inc/circular_buffer.hpp
```diff
@@ -62,7 +72,14 @@ class CircularBuffer {
      */
     uint16_t getFreeSpaceBytes();
 
-   private:
+    /**
+     * @brief Provide the size of the buffer.
+     *
+     * @return uint16_t The size of the buffer.
+     */
+    uint16_t getSize();
```
HardyYu, line 80: This function is not needed anymore

TelemetryManager/Src/TMCircularBuffer.cpp
HardyYu: You perhaps don't need `uint8_t* buf` parameter? Since you have access to the `buf_` now, you just read from buf_ in the function

HardyYu (commented):
Looks good. Please address the new comments I added and I will give approval to this pr :)

Yaremadzulynsky (approved):

Yaremadzulynsky (approved):

HardyYu (approved):
LGTM! Thanks for making the change!

PR #56, opened 2024-05-29
Added GSC functionality

Penguronik:
## Description

Implemented GSC transmit and removed the ISR as the driver already handles that.. Also renamed lastFullMessageEndIndex() to bytesUntilLastMessageEnd().

<br/>

<br/>

## Testing

<br/>

<br/>

## Documentation

*Milestone number and name:* Milestone 2

*Link to Asana task:* https://app.asana.com/0/1204699115748030/1206596856283249/f

<br/>

## Reminders

- [ ] Add reviewers to the PR

### DIFF
--- a/Drivers/common/drivers_config/inc/drivers_config.hpp
+++ b/Drivers/common/drivers_config/inc/drivers_config.hpp
@@ -17,6 +17,6 @@ extern SBUSReceiver* sbus_pointer;
 extern UARTDevice* pRFD900;
 extern CircularBuffer *rfd900_circular_buffer;
 
-#define RFD900_BUF_SIZE 280
+#define RFD900_BUF_SIZE 11200 // 40 * 280(max Mavlink message size)
 
 #endif
--- a/Drivers/common/drivers_config/inc/drivers_config.hpp
+++ b/Drivers/common/drivers_config/inc/drivers_config.hpp
@@ -17,4 +17,6 @@ extern SBUSReceiver* sbus_pointer;
 extern UARTDevice* pRFD900;
 extern CircularBuffer *rfd900_circular_buffer;
 
+#define RFD900_BUF_SIZE 280
+
 #endif
--- a/TelemetryManager/Src/GroundStationCommunication.cpp
+++ b/TelemetryManager/Src/GroundStationCommunication.cpp
@@ -23,6 +23,10 @@ GroundStationCommunication::~GroundStationCommunication() {
 void GroundStationCommunication::transmit(TMCircularBuffer &transmissionBuffer) {
     // START: Send the bytes in transmissionBuffer to the ground station via RFD900
     int bytesToTransmit = transmissionBuffer.bytesUntilLastMessageEnd();
+    
+    if (bytesToTransmit > RFD900_BUF_SIZE) {
+        bytesToTransmit = RFD900_BUF_SIZE;
+    }
 
     for (int i {0}; i < bytesToTransmit; ++i) {
         internalBuffer_[i] = transmissionBuffer.dequeue();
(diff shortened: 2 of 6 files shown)


TelemetryManager/Src/GroundStationCommunication.cpp
```diff
@@ -22,24 +22,14 @@ GroundStationCommunication::~GroundStationCommunication() {
 
 void GroundStationCommunication::transmit(TMCircularBuffer &transmissionBuffer) {
     // START: Send the bytes in transmissionBuffer to the ground station via RFD900
+    int bytesToTransmit = transmissionBuffer.bytesUntilLastMessageEnd();
+
+    for (int i {0}; i < bytesToTransmit; ++i) {
+        internalBuffer_[i] = transmissionBuffer.dequeue();
+    }
+    pRFD900->transmit(internalBuffer_, bytesToTransmit);
```
Yaremadzulynsky, line 30: Is there any logic to handle the case in which the bytes to transmit are longer than the RFD900_BUFF_SIZE? If not, are you able to brainstorm some ways of dealing with it? Thx.

Yaremadzulynsky (commented):
Everything looks rly good. Great job! Just had a question about the internal buffer. Also, are you able to remove redundant/commented out code to make it as clean as possible before we merge into main?

TelemetryManager/Src/GroundStationCommunication.cpp
Penguronik: If they are longer, would it make sense in the context of mavlink for me to break up the msg and transmit the msg in bursts?

Drivers/common/drivers_config/inc/drivers_config.hpp
```diff
@@ -17,4 +17,6 @@ extern SBUSReceiver* sbus_pointer;
 extern UARTDevice* pRFD900;
 extern CircularBuffer *rfd900_circular_buffer;
 
+#define RFD900_BUF_SIZE 280
```
HardyYu, line 20: why did we choose the maximum limit of possible mavlink messages instead of making the buffer size larger which allows it to be able to hold more messages at a time? The ring buffer should be meant to temporally hold more messages when the processor is busy doing other stuff.

HardyYu (commented):
Good stuff! I am mostly good with this pr and only left one comment there.

Drivers/common/drivers_config/inc/drivers_config.hpp
Penguronik: Last when I was talking to Yarema he mentioned that he was likely going to update this to be 10x (or at least some x) the max mavlink message but I didn't touch it since I was just moving the constant from the cpp file to the hpp file. I'm happy to make it larger right now if @Yaremadzulynsky could suggest which size to use

TelemetryManager/Src/GroundStationCommunication.cpp
Yaremadzulynsky: There shouldn't be a situation in which a mavlink message is bigger than the transmit buffer. I was more thinking the situation in which we end up overwriting the length of the internal buffer because we have lets say 100 mavlink messages each ~200 bytes in length. For now, we can probably just stop dequeuing once the internal buffer is full. Then next time we transmit, just go from where we left off. basically we just need to check if the bytesToTransmit is longer than the internal buffer and if so, just set bytes to transmit to be the length of the buffer.

Drivers/common/drivers_config/inc/drivers_config.hpp
Yaremadzulynsky: Yes. Can you make the buffer size be something like 40*max mavlink message? Thx.

HardyYu (approved):

Yaremadzulynsky (approved):

PR #58, opened 2024-07-07
Mavlink Translator

Yaremadzulynsky:
## Description

The logic to encode Mavlink messages into a byte buffer as well as decode Mavlink messages from a buffer. Also, some refactoring and editing of the main telemetry manager class.

<br/>

<br/>

## Testing

<br/>

- Encoding tests
- Decoding tests
- Encoding then decoding tests

<br/>

## Documentation

*Milestone number and name:*
Milestone 2

*Link to Asana task:* https://app.asana.com/0/1203458353737758/1206596856283252/f

<br/>

## Reminders

- [x] Add reviewers to the PR


HardyYu (approved):
Good stuff. I can't ensure that the code will work just by looking at it, but I didn't see any issue. We will figure this out when we actually run it.

Penguronik (commented):
Overall looks great, left one comment about naming and another bigger suggestion but I'm happy to approve without addressing that suggestion too

TelemetryManager/Inc/MavlinkEncoder.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include "Official_Mavlink_2_Library/common/mavlink.h"
+#include "TMCircularBuffer.hpp"
+
+/**
+ * Setting a predefined buffer size, derived from the maximum packet length allowed by MAVLink.
+ * @note IMPORTANT: IF THE THE MAX_ARRAY_BUFFER_SIZE < SIZE OF BYTES INCOMING MEMORY MAY
+ * BE OVERWRITTEN!
+ */
+#define MAX_ARRAY_BUFFER_SIZE (25 * MAVLINK_MAX_PACKET_LEN)
```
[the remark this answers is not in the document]
Yaremadzulynsky: You are correct! I don't see it used anywhere either lol. Must have been a remnant of the previous TM Mavlink Encoder I wrote. Will remove it.

TelemetryManager/Inc/MavlinkDecoder.hpp
```diff
@@ -0,0 +1,78 @@
...
+ *     //print out the message
+ *     std::cout << "ATTITUDE" << std::endl;
+ *     std::cout << "roll: " << message.roll << std::endl;
+ *     std::cout << "pitch: " << message.pitch << std::endl;
+ *     std::cout << "yaw: " << message.yaw << std::endl;
+ *     std::cout << "rollspeed: " << message.rollspeed << std::endl;
+ *     std::cout << "pitchspeed: " << message.pitchspeed << std::endl;
+ *     };
+ */
+#define REGISTER_DECODER(msgId, baseName, postProcessor)    \
```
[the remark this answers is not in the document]
Yaremadzulynsky: Great suggestion! Originally, I thought we would need to decode every type of message as it came in, so tried to make the code as dynamic as possible. Now, it's clear that we will be picking and choosing which message types should be supported. I have implemented your suggestion but have left the old code commented out in case the code ends up being too verbose in the future once we support more message types.

You clearly took the time to read & understand the code and make insightful comments, so thank you!

TelemetryManager/Src/MavlinkDecoder.cpp
```diff
@@ -0,0 +1,82 @@
...
+// must return true if the message was handled successfully
+bool handle_heartbeat(mavlink_message_t &msg) {
+    mavlink_heartbeat_t payload;
+    mavlink_msg_heartbeat_decode(&msg, &payload);
+
+    // do what you want with the payload
+    return true;
+}
+
+void MavlinkDecoder::decodeMsg(mavlink_message_t &msg, bool &isMessageDecoded) {
```
Penguronik, line 64: function looks good but noticed you dont set isMessageDecoded anywhere. Also might be worth to just return the decode status as a bool as the function return value instead?

Penguronik (changes requested):
Updates LGTM, just one concern

TelemetryManager/Src/MavlinkDecoder.cpp
Yaremadzulynsky: Yep. Makes sense. Will change that.

Yaremadzulynsky: Changed

Penguronik (approved):
LGTM!

PR #59, opened 2024-07-13
Watchdog init in the constructor

HardyYu:
Watchdog init in the constructor
## Description

The original watchdog is initiated by 'MX_IWDG_Init()', which is very confusing. Even if someone doesn't initiate a watchdog instance, they still need to refresh a watchdog. 
This PR fixed this issue by moving the initiation of the watchdog inside of the driver constructor.

## Testing

Primarily tested with some led lights

## Documentation

Maybe we should document this??? idk


Penguronik (approved):
LGTM

PR #62, opened 2024-09-25
IMU Driver for ICM42688

kelvinos2624:
Created cpp and header files for the ICM42688. The code has been structured from the tested code found at https://github.com/UWARG/efs-kitchen-sinks/blob/main/icm42688.zip

## Description

The ICM Driver (previously created, tested, and stored in https://github.com/UWARG/efs-kitchen-sinks/blob/main/icm42688.zip) has been restructured into a virtual class header file, the actual driver header file, and the driver code.

This was done to leave the option for the flight controller to initiate a different imu hardware driver or run a mock imu class for unit test purposes

## Testing

Functionality was validated with the original code on an STM board, where accelerometer and gyroscope data were successfully read from the IMU. The accuracy of accelerometer data was determined by testing for a 1g reading on each axis. The accuracy of gyroscope data was determined by temporarily implementing sensor fusion to obtain angles, from which the angles were validated.
Note: Testing for the code in this PR has not been done; only for the code in kitchensinks

N/A

## Documentation

*Milestone number and name:*
M3

*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204645027312230

N/A

### DIFF
--- a/Drivers/imu/inc/icm42688.hpp
+++ b/Drivers/imu/inc/icm42688.hpp
@@ -1,0 +1,109 @@
+// icm42688.hpp
+
+#ifndef ICM42688_HPP
+#define ICM42688_HPP
+
+#include "imu_driver.hpp"
+#include <stdint.h>
+
+class ICM42688 : public IMUDriver {
+    public:
+        /**
+        * Sets up the IMU's initial starting conditions by setting 
+        * the cs pin and calling reset(), setLowNoiseMode(), and calibrate()
+         *
+        * @return none
+         */
+        void beginMeasuring() override;
+        
+        /**
+        * Calibrates the gyroscope. Algorithm adapted from https://github.com/finani/ICM42688/tree/master/src
+         *
+        *  @return none 
+         */
+        void calibrate() override;
+        
+        /**
+        * Retrieve and process IMU accelerometer and gyroscope data into m/s and deg/s
+         *
+        * @param data_buffer -> array used to store and process raw data from IMU
+         * 
+        * @return IMUData_t structure consisting of accelerometer and gyroscope data in m/s and deg/s
+         */
+        IMUData_t getResult(uint8_t * data_buffer) override;
+
+        /**
+        * Communicate with IMU via SPI to read raw data
+         * 
+        * @param sub_address -> memory address of starting byte to be retrieved
+        * @param count -> number of bytes to be retrieved
+        * @param dest -> array to be populated with raw data
+         * 
+        * @return none
+         */
+        void readRegister(uint8_t sub_address, uint8_t count, uint8_t * dest);
+        
+        /**
+        * Communicate with IMU via SPI to write data onto IMU
+         * 
+        * @param sub_address -> memory address of byte to be written on
+        * @param data -> data to be written onto IMU
+         * 
+        * @return none
+         */
+        void writeRegister(uint8_t sub_address, uint8_t data);
+
+        /**
+        * Resets the IMU's device configurations. Necessary to initilize IMU
+         * 
+        * @return none
+         */
+        void reset();
+
+        /**
+        * Sets the IMU's power management settings to low noise mode. Necessary to initilaize IMU
+         * 
+        * @return none
+         */
+        void setLowNoiseMode();
+
+        /**
+        * Sets the IMU register address bank to a value of 0 - 4
+         *
+        * @param bank -> Bank number (0 - 4)
+         * 
+        * @return none
+         */
+        void setBank(uint8_t bank);
+
+        /**
+        * Configures the full-scale range of the gyroscope on the IMU. Adapted from https://github.com/finani/ICM42688/tree/master/src
+         * 
+        * @param fssel -> full-scale selection for the gyro
+         * 
+        * @return none
+         */
+        void setGyroFS(uint8_t fssel);
+
+    private:
+        //Constants (can be found in ICM42688 documentation https://www.cdiweb.com/datasheets/invensense/ds-000347-icm-42688-p-v1.2.pdf)
+        const uint8_t REG_BANK_SEL = 0x76;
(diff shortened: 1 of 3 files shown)


Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,137 @@
+// icm42688.cpp
+
+#include "icm42688.hpp"
+#include "stm32l5xx_hal_conf"
```
HardyYu, line 4: this line looks weird, it doesn't have a file extension

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,109 @@
...
+        /**
+        * Communicate with IMU via SPI to read raw data
+         * 
+        * @param sub_address -> memory address of starting byte to be retrieved
+        * @param count -> number of bytes to be retrieved
+        * @param dest -> array to be populated with raw data
+         * 
+        * @return none
+         */
+        void readRegister(uint8_t sub_address, uint8_t count, uint8_t * dest);
```
HardyYu, line 44: Ideally, we keep all the methods below private. Since the three methods above are the methods that users would call/interface, the functions below are those that help facilitate imu reading, not something that users want to interact with directly. We should keep them private and hide them.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#ifndef ICM42688_HPP
+#define ICM42688_HPP
+
+#include "imu_driver.hpp"
+#include <stdint.h>
+
+class ICM42688 : public IMUDriver {
+    public:
+        /**
```
HardyYu, line 11: Probably the constructor is missing here. It would be great if we could make the constructor take in an SPI instance and the CS pin. This way, this driver is not limited to only working with spi1 and cs pin at GPIOA4.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,109 @@
...
+        const uint8_t UB0_REG_TEMP_DATA1 = 0x1D;
+
+        //Variables used in gyro calibration
+        float gyro_scale = 0;
+        uint8_t current_fssel = 0;
+        uint8_t gyroFS = 0;
+        float gyroBD[3] = {0, 0, 0};
+        float gyrB[3] = {0, 0, 0};
+        float gyr[3] = {0, 0, 0};
+        uint8_t gyro_buffer[14];
```
HardyYu, line 102: The numbers 14, 7, and 7 here are magic numbers in this context. I would suggest creating macros to name these numbers, so the code would be much more readable.

I would actually also recommend replacing lines such as `const uint8_t REG_BANK_SEL = 0x76;`, to `#define REG_BANK_SEL  0x76`. In this way, these numbers won't take up any memory, they are handled by the preprocessor instead.

HardyYu (changes requested):
Good stuff! I like how you made detailed comments in the code, and clean format. They make the code super readable. Thanks for the effort you spent in this pr.

Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <string.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+#define CS_GPIO_PORT GPIOA
+#define CS_PIN GPIO_PIN_4
+
+void ICM42688::readRegister(uint8_t sub_address, uint8_t count, uint8_t * dest) {
+    //Set read bit for register address
+    uint8_t tx = sub_address | 0x80;
```
antholuo, line 17: Be careful with magic numbers

Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    IMUData_t IMUData {};
+
+    IMUData.accx = (float)raw_meas[1] / 2048.0;     //Sensitivity Scale Factor in LSB/g
+    IMUData.accy = (float)raw_meas[2] / 2048.0;     //2048.0 Corresponds with +/- 16g
+    IMUData.accz = (float)raw_meas[3] / 2048.0;
+
+    IMUData.gyrx = (float)raw_meas[4] / 16.4;       //Sensitivity Scale Factor LSB/dps
+    IMUData.gyry = (float)raw_meas[5] / 16.4;       //16.4 Corresponds with +/- 2000dps
+    IMUData.gyrz = (float)raw_meas[6] / 16.4;
```
antholuo, line 134: Are these scales selectable? Why did we choose these values?

kelvinos2624: They are selectable and I believe the values chosen are scale factors reflective of the IMU's default accel and gyro ranges (which are +/- 16g and +/- 2000dps respectively). These values prioritize a larger range over higher sensitivity. If we want more sensitivity, we would need to sacrifice range to do so. Currently, the code is not designed to adjust accelFS; if required though, it could be added. In the meantime, I'll use macros for these values for better readability and to make future changes easier.

Drivers/imu/src/icm42688.cpp
antholuo: Are you on Linux? Try building on linux, that *might* tell you where the error is.

Drivers/imu/src/icm42688.cpp
antholuo: Sounds good.

If you're doing this properly, create an ENUM (const) or defines (follow the warg style guide), and then when you call these numbers you would go SENSITIVITY.12G or SENSITIVITY.LOW.

Drivers/imu/src/icm42688.cpp
antholuo: I also just went through the build - I can't find any reference to "icm" or "hal_conf", are we sure your files are included properly in cmakelists.txt and are being built?

PR #63, opened 2024-10-09
IMU ICM42688 Driver

kelvinos2624:
## Description

The ICM Driver (previously created, tested, and stored in https://github.com/UWARG/efs-kitchen-sinks/blob/main/icm42688.zip) has been restructured into a virtual class header file, the actual driver header file, and the driver code.

This was done to leave the option for the flight controller to initiate a different imu hardware driver or run a mock imu class for unit test purposes

## Testing

Functionality was validated with the original code on an STM board, where accelerometer and gyroscope data were successfully read from the IMU. The accuracy of accelerometer data was determined by testing for a 1g reading on each axis. The accuracy of gyroscope data was determined by temporarily implementing sensor fusion to obtain angles, from which the angles were validated.
Note: Testing for the code in this PR has not been done; only for the code in kitchensinks

N/A

## Documentation

*Milestone number and name:*
M3
*Link to Asana task:*
https://app.asana.com/0/1203458353737758/1204645027312230
N/A

### DIFF
--- a/Drivers/imu/inc/icm42688.hpp
+++ b/Drivers/imu/inc/icm42688.hpp
@@ -5,9 +5,17 @@
 
 #include "imu_driver.hpp"
 #include <stdint.h>
+#include "stm32f4xx_hal.h"
+#include "stm32f4xx_hal_gpio.h"
+
+#define GYRO_CALIBRATION_DATA_BUFFER_SIZE 14
+#define GYRO_CALIBRATION_RAW_DATA_BUFFER_SIZE 7
+#define RAW_DATA_BUFFER_SIZE 7
 
 class ICM42688 : public IMUDriver {
     public:
+        ICM42688(SPI_HandleTypeDef * SPI_HANDLE, GPIO_TypeDef * CS_GPIO_PORT, uint16_t CS_PIN);
+
         /**
         * Sets up the IMU's initial starting conditions by setting 
         * the cs pin and calling reset(), setLowNoiseMode(), and calibrate()
@@ -32,6 +40,7 @@ class ICM42688 : public IMUDriver {
          */
         IMUData_t getResult(uint8_t * data_buffer) override;
 
+    private:
         /**
         * Communicate with IMU via SPI to read raw data
          * 
@@ -85,7 +94,6 @@ class ICM42688 : public IMUDriver {
          */
         void setGyroFS(uint8_t fssel);
 
-    private:
         //Constants (can be found in ICM42688 documentation https://www.cdiweb.com/datasheets/invensense/ds-000347-icm-42688-p-v1.2.pdf)
         const uint8_t REG_BANK_SEL = 0x76;
         const uint8_t UB0_REG_DEVICE_CONFIG = 0x11;
@@ -99,11 +107,15 @@ class ICM42688 : public IMUDriver {
         float gyroBD[3] = {0, 0, 0};
         float gyrB[3] = {0, 0, 0};
         float gyr[3] = {0, 0, 0};
-        uint8_t gyro_buffer[14];
-        int16_t raw_meas_gyro[7];
+        uint8_t gyro_buffer[GYRO_CALIBRATION_DATA_BUFFER_SIZE];
+        int16_t raw_meas_gyro[GYRO_CALIBRATION_RAW_DATA_BUFFER_SIZE];
 
         //Used to hold raw IMU data
-        int16_t raw_meas[7];
+        int16_t raw_meas[RAW_DATA_BUFFER_SIZE];
+
+        SPI_HandleTypeDef * SPI_HANDLE;
+        GPIO_TypeDef * CS_GPIO_PORT;
+        uint16_t CS_PIN;
 };
 
 #endif //ICM42688_HPP
--- a/Drivers/imu/inc/icm42688.hpp
+++ b/Drivers/imu/inc/icm42688.hpp
@@ -1,0 +1,109 @@
+// icm42688.hpp
+
+#ifndef ICM42688_HPP
+#define ICM42688_HPP
+
+#include "imu_driver.hpp"
+#include <stdint.h>
+
+class ICM42688 : public IMUDriver {
+    public:
+        /**
+        * Sets up the IMU's initial starting conditions by setting 
+        * the cs pin and calling reset(), setLowNoiseMode(), and calibrate()
+         *
+        * @return none
+         */
+        void beginMeasuring() override;
+        
+        /**
+        * Calibrates the gyroscope. Algorithm adapted from https://github.com/finani/ICM42688/tree/master/src
+         *
+        *  @return none 
+         */
+        void calibrate() override;
+        
+        /**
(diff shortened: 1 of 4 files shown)


Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,122 @@
...
+        /**
+        * Communicate with IMU via SPI to read raw data
+         * 
+        * @param sub_address -> memory address of starting byte to be retrieved
+        * @param count -> number of bytes to be retrieved
+        * @param dest -> array to be populated with raw data
+         * 
+        * @return none
+         */
+        void readRegister(uint8_t sub_address, uint8_t count, uint8_t * dest);
```
antholuo, line 54: nit: `count` is not as descriptive as `num_bytes_to_retrieve`. It's not toooo much longer.

Also, sub_address and dest, try and use concatenation either or don't use it at all.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "imu_driver.hpp"
+#include <stdint.h>
+#include "stm32l5xx_hal.h"
+#include "stm32l5xx_hal_gpio.h"
+#include <stm32l552xx.h>
+
+#define GYRO_CALIBRATION_DATA_BUFFER_SIZE 14
+#define GYRO_CALIBRATION_RAW_DATA_BUFFER_SIZE 7
+#define RAW_DATA_BUFFER_SIZE 7
```
antholuo, line 14: Any explanation for why these are the way they are?

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,122 @@
...
+        
+        /**
+        * Communicate with IMU via SPI to write data onto IMU
+         * 
+        * @param sub_address -> memory address of byte to be written on
+        * @param data -> data to be written onto IMU
+         * 
+        * @return none
+         */
+        void writeRegister(uint8_t sub_address, uint8_t data);
```
antholuo, line 64: nit again, `data_to_imu` is more descriptive.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,122 @@
...
+         * 
+        * @return none
+         */
+        void setGyroFS(uint8_t fssel);
+
+        //Constants (can be found in ICM42688 documentation https://www.cdiweb.com/datasheets/invensense/ds-000347-icm-42688-p-v1.2.pdf)
+        const uint8_t REG_BANK_SEL = 0x76;
+        const uint8_t UB0_REG_DEVICE_CONFIG = 0x11;
+        const uint8_t UB0_REG_PWR_MGMT0 = 0x4E;
+        const uint8_t UB0_REG_TEMP_DATA1 = 0x1D;
```
antholuo, line 102: Consider using constexpr? or use defines, as you were using up above.

Either way, you should be consistent in your usage of macros or not-macros. Not that it matters much.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+        //Variables used in gyro calibration
+        float gyro_scale = 0;
+        uint8_t current_fssel = 0;
```
antholuo, line 106: fs_sel?

Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "gpio.h"
+#include <string.h>
+#include <stdio.h>
+
+#define REG_BANK_SEL 0x76
+#define UB0_REG_DEVICE_CONFIG 0x11
+#define UB0_REG_PWR_MGMT0 0x4E
+#define UB0_REG_TEMP_DATA1 0x1D
+
+#define BIT_READ 0x80
```
antholuo, line 17: not super descriptive - hard to tell what this is doing or going to do

Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,163 @@
...
+    writeRegister(0x4F, reg);
+    
+    gyro_scale = (2000.0f / (float)(1 << fssel)) / 32768.0f;
+    gyroFS = fssel;
+}
+
+void ICM42688::calibrate() {
+    //Set at a lower range (more resolution since IMU not moving)
+    const uint8_t current_fssel = gyroFS;
+    setGyroFS(0x03);        //Set to 250 dps
```
antholuo, line 112: magic number 0x03. Should this be a #define or constexpr?

Drivers/imu/src/icm42688.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    //Take samples and find bias
+    gyroBD[0] = 0;
+    gyroBD[1] = 0;
+    gyroBD[2] = 0;
+
+    for (size_t i = 0; i < 3; i++) {
+        getResult(gyro_buffer);
+        for (size_t i = 0; i < 7; i++) {
+            raw_meas_gyro[i] = ((int16_t)gyro_buffer[i * 2] << 8) | gyro_buffer[i * 2 + 1];
```
antholuo, line 122: Math explanation might be nice in a comment.

antholuo (approved):
Looks good, only thing I have are small nits.

Maybe squash the commits into one descriptive commit message using rebase -i.

Also, try and include some screenshots of the test data you received in the PR desc, if you have any.

Drivers/imu/inc/icm42688.hpp
```diff
@@ -0,0 +1,115 @@
...
+        float gyr[3] = {0, 0, 0};
+        uint8_t gyro_buffer[GYRO_CALIBRATION_DATA_BUFFER_SIZE];
+        int16_t raw_meas_gyro[RAW_MEAS_BUFFER_SIZE];
+
+        //Used to hold raw IMU data
+        int16_t raw_meas[RAW_MEAS_BUFFER_SIZE];
+
+        SPI_HandleTypeDef * SPI_HANDLE;
+        GPIO_TypeDef * CS_GPIO_PORT;
+        uint16_t CS_PIN;
```
HardyYu, line 112: It will be the best if you can change these member variables according to https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/2268692486/Style+Standards . Having only these member variables all caps are weird. And also conventionally, ppl think all caps means macro.

HardyYu (commented):
Fantastic work Kelvin! I only left a minor request for change. Sorry for the late review.

HardyYu (approved):
Thank you for making the change Kelvin!

Drivers/imu/inc/icm42688.hpp
kelvinos2624: I have renamed it in the latest commit

Drivers/imu/src/icm42688.cpp
kelvinos2624: Ah thanks, you're right; I changed it to be a #define in the latest commit.

PR #65, opened 2024-10-27
Dependency injection

Yaremadzulynsky:
## Description

All dependency objects used by TelemetryManager (e.g. GroundStationCommunication) are now created in SystemManager before the TM object. Dependencies are passed directly to the TM constructor.

All state data pointers used by TM have been organized to a struct.

<br/>

TM can now be created from mock instances for testing.
The order of object construction can be changed, so we can wait until we have all dependencies ready before initializing a TM object.
TM objects can potentially be made outside of SystemManager 

<br/>

## Testing

- tools.ps1 compile test

<br/>

## Documentation

*Milestone number and name:*
M3

*Link to Asana task:* https://app.asana.com/0/1203458353737758/1208312402823490/f

## Reminders

- [x] Add reviewers to the PR


SystemManager/Src/SystemManager.cpp
Yaremadzulynsky, line 1: Awesome. One note I have is that the SM constructor is starting to look a little cluttered with TM stuff. What we could do is make a function in the SM cpp file called something like setupTM() which will do the TM setup stuff, then call tm->init() after everything is set up.

TelemetryManager/Inc/TelemetryManager.hpp
Yaremadzulynsky, line 1: Looking great. As you mentioned, when we switch from references to pointers, we no longer need to assert that the pointer has been initialized. This could lead to a potential hard fault during flight (would be very bad) if we try to dereference a pointer that doesn't exist. Can we somehow check if the pointer has been initialized before we dereference it?

SystemManager/Src/SystemManager.cpp
Yaremadzulynsky, line 1: Awesome. One note I have is that the SM constructor is starting to look a little cluttered with TM stuff. What we could do is make a function in the SM cpp file called something like setupTM() which will do the TM setup stuff, then call tm->init() after everything is set up.

TelemetryManager/Src/TelemetryManager.cpp
Yaremadzulynsky, line 1: The constructor looks infinitely cleaner! Nice job!

Yaremadzulynsky (commented):
Great work!

Yaremadzulynsky (commented):
LGTM

larry-pan (approved):
Looks good

